Issue #31, opened 2015-08-03
WF32 and WiFIRE do not build Blink


KeithV:
Get errors when either the WF32 or WiFIRE builds the example blink sketch

ricklon:
Discovered two issues. 

1. Both boards include .c/.cpp files inside of Board_Data.c. This is compilation error. This needs to be done a different way. All .c and .cpp files in the variant folder is now compiled. THis is different than MPIDE. This will be added back in MPIDE.

2. build.ldscript is used to find the ldscripts. Currently, testing changing the compilation recipes to make a valid ld feature.

EmbeddedMan:
Rick,

Are you going to be making the changes to the Board_Data.c files? Or do you
want me or Keith to do it?

*Brian

ricklon:
I'd like you or Keith to make the changes. I'm chasing down the next set of errors, but I think they are related to the testing that eliminated the error includes.

MPIDE
```
/Applications/mpide.app/Contents/Resources/Java/hardware/pic32/compiler/pic32-tools/bin/pic32-g++  -Os  -Wl,--gc-sections  -mdebugger  -mno-peripheral-libs  -nostartfiles  -mprocessor=32MZ2048ECG100  -o  /var/folders/bx/zg7wx2513knbft_4ch_p5m6c0000gp/T/build611625737197987005.tmp/wf32ex2serial.cpp.elf  /var/folders/bx/zg7wx2513knbft_4ch_p5m6c0000gp/T/build611625737197987005.tmp/wf32ex2serial.cpp.o    /var/folders/bx/zg7wx2513knbft_4ch_p5m6c0000gp/T/build611625737197987005.tmp/core.a  -L/var/folders/bx/zg7wx2513knbft_4ch_p5m6c0000gp/T/build611625737197987005.tmp  -lm  -T  /Applications/mpide.app/Contents/Resources/Java/hardware/pic32/variants/WiFire/MZ-application-32MZ2048ECX.ld  -T/Applications/mpide.app/Contents/Resources/Java/hardware/pic32/cores/pic32/chipKIT-application-COMMON-MZ.ld
```

Arduino
```
[...]

PR #39, opened 2015-08-22
Partial fix for issue #31 - removed two #include statements from WiFi…

EmbeddedMan:
…re/Board_Data.c that were including other .c files.


ricklon:
The next problem. Those files that are now not include are compiled and then fail. After those files are deleted the functionality they provide fails. :-(

EmbeddedMan:
Crud. Yeah, those files need to be compiled in, as their functionality
implements the ADC stuff that Keith worked on. This might be something
Keith should look at.

PR #48, opened 2015-08-28
Added PGM_P pointer type to AVR pgmspace compatibility

majenkotech:
Came across an instance where `PGM_P` was being used to point to a PROGMEM string.  We lacked that alias.


ricklon:
Thanks for finding that alias. We'll merge.

PR #49, opened 2015-08-28
Added delay stubs to alias custom AVR delay functions

majenkotech:
The functions `_delay_ms()` and `_delay_us()` are defined by avr-libc and some library authors use them when they really shouldn't (probably when porting code from other sources).  By providing these aliases to the standard API functions we allow these libraries to be compiled without changes.

Other cores (like teensy 3 and maple) also provide this aliasing for the same reason.


adamwolf:
Why @chipkitbot why?!

majenkotech:
It's @chipkitb0rk!

PR #50, opened 2015-08-30
Added missing { from Client.cpp

majenkotech:


majenkotech:
Well, @chipkitbot, that almost worked that time.  It came up success for a bit, and then went to fail. So it worked but you don't know it worked?

majenkotech:
Could it be because chipkitjenkins.wayneandlayne.com can't be resolved, so you can't get the results from the server?

PR #70, opened 2015-09-13
Ported _doprnt() from RetroBSD for Print class

majenkotech:
This allows you to use printf() as a member function on any Print based object, such as:

    Serial.printf("The temperature is: %d and the time is: %02d:%02d:%02d\r\n",
        temperature, hours, minutes, seconds);

### DIFF
--- a/pic32/cores/pic32/Print.cpp
+++ b/pic32/cores/pic32/Print.cpp
@@ -243,3 +243,13 @@ size_t Print::printFloat(double number, uint8_t digits)
   
   return n;
 }
+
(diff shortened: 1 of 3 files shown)


majenkotech:
Sketch impact of using it, looking at objdump:

    00001e94 _Z7_doprntPKcPvP5Print

That's about 7828 in decimal, but the actual size increase is somewhat more due to the Print bindings.  Overall it seems to be about 15K more than just using Serial.print() and about 14K smaller than using sprintf().

Issue #68, opened 2015-09-13
Dependency check on recompilation not performed


ricklon:
Dependency check on recompilation not performed this is because the dependency files were not generated. 

Add the following to platform.txt and it generates properly.
compiler.cpp.extra_flags=-MMD
compiler.c.extra_flags=-MMD

If this value is changed in boards.txt dependencies will not be generated for a specific board.

EmbeddedMan:
Rick, I'm unable to verify that your PR fixed this issue. For an empty sketch, I get the same compile time first time vs. repeated compiles with no changes. (This is with latest chipKIT-core built from Github repo contents).

Did you test this and confirm that it fixed the problem on your system?

ricklon:
I'm double checking now. What board did you test with?

--Rick

On Tue, Sep 15, 2015 at 6:17 PM, Brian Schmalz <notifications@github.com>
wrote:

> Rick, I'm unable to verify that your PR fixed this issue. For an empty
> sketch, I get the same compile time first time vs. repeated compiles with
> no changes. (This is with latest chipKIT-core built from Github repo
> contents).
>
> Did you test this and confirm that it fixed the problem on your system?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/chipKIT32/chipKIT-core/issues/68#issuecomment-140562032>
> .
>



-- 
Co-founder
Fair Use Building and Research (FUBAR) Labs
http://fubarlabs.org

EmbeddedMan:
Fubarino Mini.

ricklon:
I tested in the latest Arduino 1.6.6 master environment.

It didn't work in the Arduino  1.6.5.  The flag is included in the
compilation for both. Something has to be wrong about the comparison.

We need to file a new issue that it's not working in Arduino 1.6.5.

--Rick

On Tue, Sep 15, 2015 at 6:45 PM, Brian Schmalz <notifications@github.com>
wrote:

> Fubarino Mini.
>
> On Tue, Sep 15, 2015 at 5:44 PM, Ricklon <notifications@github.com> wrote:
>
> > I'm double checking now. What board did you test with?
> >
> > --Rick
> >
> > On Tue, Sep 15, 2015 at 6:17 PM, Brian Schmalz <notifications@github.com
> >
> > wrote:
> >
> > > Rick, I'm unable to verify that your PR fixed this issue. For an empty
> > > sketch, I get the same compile time first time vs. repeated compiles
> with
[...]

PR #72, opened 2015-09-16
Fix for issue #68 : dependency check. The option needs to be -MMD, bu…

EmbeddedMan:
…t it was previously spelled -MDD, thus it didn't work very well. This fixes that.


EmbeddedMan:
Rick, if you could confirm that this fix in fact works for you. (It does for me.)

Issue #67, opened 2015-09-13
Everyboard in boards.txt overides default compiler.c.flags, and compiler.cpp.flags


ricklon:
Every board in boards.txt overrides default compiler.c.flags, and compiler.cpp.flags even when they might not need to.

EmbeddedMan:
What's the task to do here? To go through each compiler.cpp.flags assignment in every variant and remove it if it's the same as he master copy in boards.txt?

ricklon:
Correct. If the board relies in default settings or needs custom settings.

--Rick

EmbeddedMan:
So for an option like this, in an individual board.txt file, there's now
way to specify "keep the existing value from the global boards.txt file,
but add my own little customization"? That's too bad.

I think I can take care of this one too.

*Brian

ricklon:
I think there is an equivalent of append, but I need to confirm Arduino
implemented it.

-_Rick

On Thu, Sep 17, 2015 at 10:46 AM, Brian Schmalz <notifications@github.com>
wrote:

> So for an option like this, in an individual board.txt file, there's now
> way to specify "keep the existing value from the global boards.txt file,
> but add my own little customization"? That's too bad.
>
> I think I can take care of this one too.
>
> *Brian
>
> On Thu, Sep 17, 2015 at 9:43 AM, Ricklon <notifications@github.com> wrote:
>
> > Correct. If the board relies in default settings or needs custom
> settings.
> >
> > --Rick
> >
> > On Thu, Sep 17, 2015, 10:08 AM Brian Schmalz <notifications@github.com>
> > wrote:
> >
> > > What's the task to do here? To go through each compiler.cpp.flags
> > > assignment in every variant and remove it if it's the same as he master
> > > copy in boards.txt?
> > >
> > > —
> > > Reply to this email directly or view it on GitHub
> > > <
> >
> https://github.com/chipKIT32/chipKIT-core/issues/67#issuecomment-141098218
> > >
> >
> > > .
> > >
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <
> https://github.com/chipKIT32/chipKIT-core/issues/67#issuecomment-141108833
> >
> > .
>
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/chipKIT32/chipKIT-core/issues/67#issuecomment-141110198>
> .
>



-- 
Co-founder
Fair Use Building and Research (FUBAR) Labs
http://fubarlabs.org

[...]

PR #76, opened 2015-09-18
Fix for issue #67: Every entry in boards.txt overrides platform.txt f…

EmbeddedMan:
…or compiler.c.flags. This has now been fixed, and only platform.txt contains compiler.c.flags. (Same for compiler.cpp.flags)

I've done some limited testing (making sure that a sample sketch compiles on all boards, and runs on a representative sample of boards) and this change does not appear to break anything. Notice however that we are using the -G1024 command line option of all boards, not just the ones with limited Flash. Based on the comments in the boards.txt file, I'm not sure if this is always needed or not.

Another change is -g rather than -g3. It's not clear to me what -g3 does compared with -g.

Another change is that we don't use the -fframe-base-loclist option.

Note that none of these are 'changes', since this PR doesn't actually cause our compiler to have any different options than it did before. This is because every board (in the boards.txt file)  was over-riding the default values in platform.txt. Now the value in platform.txt is what was in boards.txt for each board, but we've also removed all of the (identical) option settings in the boards.txt file.

### DIFF
--- a/pic32/boards.txt
+++ b/pic32/boards.txt
@@ -79,12 +79,6 @@ cerebot_mx3ck.ccflags=ffff
 cerebot_mx3ck.ldscript=chipKIT-application-32MX320F128.ld
 # end of new items
 
-# Use a high -Gnum for devices that have less than 64K of data memory
-# For -G1024, objects 1024 bytes or smaller will be accessed by
-# gp-relative addressing
-cerebot_mx3ck.compiler.c.flags=-O2 -c -mno-smart-io -w -ffunction-sections -fdata-sections -G1024 -g -mdebugger -Wcast-align -fno-short-double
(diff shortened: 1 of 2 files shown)


adamwolf:
Look at that!  All checks have passed!  (You can't click on the link to get to chipkitjenkins.wayneandlayne.com yet, but it looks like the build issues previous were a polling interval issue.)

EmbeddedMan:
Yes, that is very good. I didn't get any sort of an e-mail from chipkitbot
asking to do the build though. Is that supposed to happen? And how are we
to know if the build is done? Does that all depend on getting
chipkitjenkins opened up?

*Brian

adamwolf:
You're a trusted person, so it'll auto build your PRs. If you'd rather have it ask, I can take you off the list.

I think it posts "Checks are pending", then "all checks have passed".

EmbeddedMan:
Ahh, OK. No need to change - I like being trusted. :-) Just making sure I understand what's going on.

So when it says "all checks have passed" that means there's now an image ready waiting for me to download and test?

adamwolf:
Sorry, I commented as chipkitbot by accident.

PR #84, opened 2015-09-30
put timing in SD, created DTWI library, added I2C repeated start

KeithV:
Split DTWI out into its own library
Added repeated start to Wire
removed nops from SD, added real timing
restored WiFiShield SPI conflict code in SD
fixed MZ linker script to add debug RAM for FPU
Allowed for more than 8 PWMs (9 on MZ) in wiring_analog.c
Removed plib.h from dp32


EmbeddedMan:
ok to test

EmbeddedMan:
No problem. I'm reviewing the PR right now, and will merge it shortly and
do my own test build for Keith.

Thanks for trying chipkitbot. I know this is really hard. We won't give up
on you!

*Brian

EmbeddedMan:
Keith, Before we merge this in, could you add some appropriate comments into Sd2Card.cpp around the code you put back in that handles the WiFi shield interrupt stuff? I can't think of an (easy) better solution, so at least there should be some solid comments describing why that code needs to be in there.

adamwolf:
ok to test

adamwolf:
Look at that!  @chipkitbot, you're getting good!

KeithV:
I think I got the comment in there. I think it put it on my original pull request. Not a Git expert, but it certainly is in my fork.

PR #116, opened 2015-12-02
Added support for board-defined TX and RX LEDs for USB serial

majenkotech:
This makes it possible for a board to have a TX and RX LED wired in for USB serial activity to be displayed.

It relies on two definitions in the Board_Defs.h file: PIN_TX_LED and PIN_RX_LED which point to a digital pin number to use as activity LEDs. You can define one or both of those and they are independent.


EmbeddedMan:
What a cool idea Matt. Love it.

Issue #145, opened 2016-01-06
Copy startup code from chipKIT32-MAX repo


jasonkajita:
Copy startup code from https://github.com/chipKIT32/chipKIT32-MAX/commit/35a91c791478a1b98bdd3610c6c199444df22ed4

EmbeddedMan:
This needs to be confirmed by somebody (Keith) that can properly test that this change is working as intended.

EmbeddedMan:
Keith, have you confirmed that the startup code you need is now included properly?

KeithV:
Looks good you can close

PR #146, opened 2016-01-06
Copy startup code from https://github.com/chipKIT32/chipKIT32-MAX/com…

jasonkajita:
Issue https://github.com/chipKIT32/chipKIT-core/issues/145


EmbeddedMan:
ok to test

EmbeddedMan:
Adam- this didn't recognize my "ok to test" even though I didn't use a CR at the end. :-( So chipkitbot still doesn't like to always listen to me.

PR #148, opened 2016-01-07
modified boards.txt so that the Quick240 does not overwrite the compi…

mskoczen:
…ler defines

removed unused Spi defines Quick240
Fixed Dspi defines for Quick240
Added defines for the Ethernet library


ricklon:
ok to test

Issue #158, opened 2016-01-22
Fix WiFire ADC resolution issue in EFADC.c


EmbeddedMan:
As reported by Dave Barringer:

The correct answer is

    ADC0TIMEbits.SELRES     = ADCCON1bits.SELRES;               // ADC0 resolution is 12 bits

(then a few lines further)

    ADC1TIMEbits.SELRES     = ADCCON1bits.SELRES;

etc (5 changes total).

EmbeddedMan:
PR reviewed and accepted as correct.

PR #159, opened 2016-01-22
Fix for chipKIT-core issue #158 : Fix WiFire ADC resolution issue in …

EmbeddedMan:
…EFADC.c


EmbeddedMan:
Confirmed by Keith and Dave.

Issue #163, opened 2016-01-31
Replace register names with base address defines in MZ variant files


EmbeddedMan:
Currently the MZ parts use lines like this in Board_Defs.h
#define	_SPI_BASE		((uint32_t) &SPI2CON)

The reason is because previous versions of the compiler did not properly define _SPI2_BASE_ADDRESS. It appears that the latest compiler does, so these should all be updated to use the proper _SPI2_BASE_ADDRESS (or whatever) so they're the same as all the other boards.

Issue #144, opened 2016-01-06
Remove instances of SPI define in all board variant files


EmbeddedMan:
They should all use DSPI0 now rather than the old Arduino SPI.

EmbeddedMan:
I need to find more information out about exactly what this issue means. Pontech found this with their Quick240 board, and it was messing them up until they went to all DSPI definitions. So now there is no more _SPI_BASE, etc. defines in the Quick240 Board_Defs.h file. Should all chipKIT variants have all of there _SPI_BASE defines completely removed?

EmbeddedMan:
Keith, Maybe you can help shed some light on this one for me. I'm not asking for you to do anything other than see if this makes sense. Pontech needed to switch over the defines in their variant file from using "SPI" to using 'DSPI0" in order to get SD card access working (I think it was SD card access) on their Quick240 board with the v1.1.0 release of core. Should this change be propagated through all other variant files? Should SPI be completely removed in favor of DSPI? Maybe that's what you've been doing already - I just want your thoughts.

EmbeddedMan:
See Jacob's issue https://github.com/chipKIT32/chipKIT-core/issues/121 for other details

KeithV:
Brian,

 

I am not sure. I don’t use, and typically have not used SPI. I think Matt at some time started to switch things over to the DSPIx #defines; I forget the reason though. I can take a look and see if I can figure it out but since I don’t use the SPI library, I am not sure exactly how that works.

 

KeithV

EmbeddedMan:
Keith, OK, cool. No problem. I'll investigate the matter further to try and understand where this came from, and what the ramifications would be to the other boards. Thanks for your help.

KeithV:
Brian,

 

I took a look and I can speculate on what Matt was trying to do.

 

In the original variants there were things like:

 



 

That was put in to support the constructor of the Arduino SPI

 



 

For NON-PPS MCU, really only the Base was needed, and originally things like:

 

_SPI2_BASE_ADDRESS

 

Where used. But then comes PPS and we need the pins, originally the const above were used, but I think Matt noticed that we were adding pins for ALL of the DSPIx classes be it needed that for PPS as well..

 

 



 

But then the compiler goofed and did not add the SPI base address, and for DSPI I fixed the problem by putting in the address of the first entry in the SPI sfrs. It was an oversite in the pic32xxxx.h file to not include the base address; but I couldn’t fix that, so this is what I did.

 



 

[...]

Issue #121, opened 2015-12-22
SPI not working in chipkit-core-windows.16778039.master-v1.0.0-74-g9cd95dd


JacobChrist:
I have a sketch that I'm now testing in both mpide-20140313 and the latest chipkit-core that talks to a Maxim6675.  The sketch works fine in mpide-20140313, however, SPI.transfer(0x00) returns either all zeros or all ones in chipkit-core.

Here is some really simple code, you probably don't need the MAX6675 to test, you can just tie SDI high or low.  If its low the output with be 0.0000, if its high it will be -100.0000.

```C
//Arduino 1.6.7 w/ chipkit-core-windows.16778039.master-v1.0.0-74-g9cd95dd

#include <SPI.h>

#define C4IO0 23
#define C4IO1 39
#define C5IO0 78
#define C5IO1 79

// Thermocouple 0/1
#define TC0_CS    C4IO0
#define TC1_CS    C4IO1
#define TC2_CS    C5IO0
#define TC3_CS    C5IO1
#define DEBUGLED  PIN_LED1

void setup() {
  Serial.begin(115200);

  pinMode(TC0_CS,OUTPUT);
  pinMode(TC1_CS,OUTPUT);
  pinMode(TC2_CS,OUTPUT);
  pinMode(TC3_CS,OUTPUT);
  digitalWrite(TC0_CS,HIGH);
  digitalWrite(TC1_CS,HIGH);
  digitalWrite(TC2_CS,HIGH);
  digitalWrite(TC3_CS,HIGH);

  pinMode(PIN_LED1,OUTPUT);
  digitalWrite(PIN_LED1,LOW);

  SPI.begin();
}

void loop() {
  static uint32_t heartbeattimelast_ms = 0;
  static bool heartbeattimelast_bool = false;
  uint32_t loopstart_ms = millis();
  if (CompareTime(heartbeattimelast_ms, loopstart_ms) >= 1000)
  {
    //Serial.println("beat");
    heartbeattimelast_ms = heartbeattimelast_ms + 1000;
    heartbeattimelast_bool = !heartbeattimelast_bool;
    digitalWrite(DEBUGLED,heartbeattimelast_bool);
    Serial.println(MAX6675readCelsius(TC0_CS), 4);
    Serial.println(MAX6675readCelsius(TC1_CS), 4);
    Serial.println(MAX6675readCelsius(TC2_CS), 4);
    Serial.println(MAX6675readCelsius(TC3_CS), 4);
[...]

PR #164, opened 2016-01-31
Fix for issue #144, #121, and #163

EmbeddedMan:
Removed all SPI definitions from variant files (we now use DSPI defines in SPI.cpp)
Updated #ifdef in several files to use __PIC32_PPS__ rather than separate CPU defines
Replaced bare register names in variant defines with _BASE_ADDRESS_ defines from compiler .h files now that they are there.


EmbeddedMan:
After being reviewed by Keith and Matt, this looks good to merge.

PR #168, opened 2016-02-11
Added option to select default DSPI port for use with SPI library

majenkotech:
This basically allows you to define the macro _SPI_PORT in the board definition, and the SPI library then uses the details of that DSPI channel instead of the default DSPI0.

If nothing is defined it just works as it always did - using DSPI0. However, if you define something like:

    #define _SPI_PORT 2

in the Board_Defs.h file it will instead use the details of DSPI2 for the SPI library.

### DIFF
--- a/pic32/libraries/SPI/SPI.cpp
+++ b/pic32/libraries/SPI/SPI.cpp
@@ -13,10 +13,52 @@
 
 #include "SPI.h"
 
-#if defined(__PIC32_PPS__)
-SPIClass SPI(_DSPI0_BASE, _DSPI0_MISO_PIN, _DSPI0_MOSI_PIN, _DSPI0_MISO_IN, _DSPI0_MOSI_OUT);
+#ifndef _SPI_PORT
+    #define _SPI_PORT 0
+#endif
+
+#if (_SPI_PORT == 0)
+    #if defined(__PIC32_PPS__)
+        SPIClass SPI(_DSPI0_BASE, _DSPI0_MISO_PIN, _DSPI0_MOSI_PIN, _DSPI0_MISO_IN, _DSPI0_MOSI_OUT);
+    #else
+        SPIClass SPI(_DSPI0_BASE);
+    #endif
+#elif (_SPI_PORT == 1)
+    #if defined(__PIC32_PPS__)
+        SPIClass SPI(_DSPI1_BASE, _DSPI1_MISO_PIN, _DSPI1_MOSI_PIN, _DSPI1_MISO_IN, _DSPI1_MOSI_OUT);
+    #else
+        SPIClass SPI(_DSPI1_BASE);
+    #endif
+#elif (_SPI_PORT == 2)
+    #if defined(__PIC32_PPS__)
+        SPIClass SPI(_DSPI2_BASE, _DSPI2_MISO_PIN, _DSPI2_MOSI_PIN, _DSPI2_MISO_IN, _DSPI2_MOSI_OUT);
+    #else
+        SPIClass SPI(_DSPI2_BASE);
+    #endif
+#elif (_SPI_PORT == 3)
+    #if defined(__PIC32_PPS__)
+        SPIClass SPI(_DSPI3_BASE, _DSPI3_MISO_PIN, _DSPI3_MOSI_PIN, _DSPI3_MISO_IN, _DSPI3_MOSI_OUT);
+    #else
+        SPIClass SPI(_DSPI3_BASE);
+    #endif
+#elif (_SPI_PORT == 4)
+    #if defined(__PIC32_PPS__)
+        SPIClass SPI(_DSPI4_BASE, _DSPI4_MISO_PIN, _DSPI4_MOSI_PIN, _DSPI4_MISO_IN, _DSPI4_MOSI_OUT);
+    #else
+        SPIClass SPI(_DSPI4_BASE);
+    #endif
+#elif (_SPI_PORT == 5)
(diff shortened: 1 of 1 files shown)


EmbeddedMan:
Matt, thanks for this. Keith recently made a very similar (in intent - not in function) PR. I'd like Keith to give his opinion on this PR as well, considering his recent work in this area.

majenkotech:
If that is commit pr #166 then that was just a single tweak to a single board to bypass the issue. This is a method by which any board can specify that SPI is not the same as DSPI0 but should be the same as DSPI<whatever>.  Keith's PR is not a fix for the problem, only a change to the MX7 board so it works with the assumption that SPI is DSPI0.  It's a completely different thing.

majenkotech:
Incidentally, if you do your board configurations right, in UECIDE you can even get the option to select your SPI port destination for yourself:

image

EmbeddedMan:
Matt - yeah, I understand yours is more generic. I still think it's trying to solve the same problem though - how to have SPI be assigned something other than DSPI0. Keith's code just fixes it in the one board where we want SPI to be something other than DSPI0, yours makes it so that any board can use any DSPI port as the default. In any case, I like the generic nature of your code a lot, but I do want to know what Keith thinks of it too.

majenkotech:
Actually, Keith's doesn't even attempt to fix the problem for the board, it just re-maps which SPI ports are on which DSPI objects. It still has SPI on DSPI0, he just changed which pins DSPI0 points to. That operation, and this PR are two completely different things. Yes, they overlap in the fact that they are to do with DSPI0 pointing to SPI, but they are completely different things entirely. 

Yes, this PR could obsolete Keith's change if he so wished, in that he could revert the DSPI objects to their former SPI channels and set _SPI_PORT, but that doesn't *have* to happen since it can stay that SPI is on DSPI0 since that is what his tweak to the board settings has ended up with.

Keith's change does not address the issue at all. It just fixes a problem in a specific board where there has historically been a mismatch between SPI and DSPI0.

EmbeddedMan:
Actually, Keith's PR was a fix for the SPI changes you made last year that
broke some of Digilent's board's SPI functionality. (The problem was
substituting DSPI0 for SPI, where some boards had SPI defined as other than
the same as DSPI0). There was no historic mis-match between SPI and DSPI0
until that change that you made. From that point on there was. Keith's PR
fixed that, and your new PR fixes it in an even better way.

I agree that your PR could obsolete Keith's PR if he wanted to (and,
considering your change, that feels like a cleaner way to handle the
issue), and I guess we'll have to agree to disagree that your PR is related
to his PR.

In any case, I don't think any of this discussion contributes to the case
for accepting this PR. The code seems like a good idea to me, and I just
wanted Keith to take a look to see if he notices anything that I've missed
since he'll be much better at seeing problems since he was recently
thinking about this same area of code because of his recent PR. That's all.
I wasn't trying to pick a fight.

majenkotech:
By mismatch I meant that historically DSPI0 and SPI both pointed to different places in that one specific board, which is why that board broke with the SPI library changes when SPI was forced to point to DSPI0. Keith's PR is a complete irrelevancy and doesn't have any bearing on this PR at all.

I agree that more eyes are always better.

KeithV:
In effect Matt's suggestion just "restores" most of the original functionality that was removed when SPI was changed to use DSPI0. Originally we had a bunch of SPI #defines that could be defined to any SPI hardware, then Matt change SPI to use DSPI0, thus breaking Pontech and Digilent MX7 boards. By  putting in _SPI_PORT, in effect you are just recreating the original defines, just mapping them to a DSPI set of defines. 

I would prefer to  either go back to the original method, OR use Matt's PR. Both restore the decoupling of DSPI0 to SPI and that is a good thing. In fact I would back out my PR on the DSPI0/DSPI2 switch and set _SPI_PORT 2 in the MX7s; there are people who assume which DSPIx is  used on which Pmod and that I just broke that with my PR. BUT.... what ever you do, make sure Pontech works (please do some testing!!!! Don't just assume....) Also this has a restriction that there IS a DSPIx defined that will be used for SPI (not necessary with the original implementation). My order of preferences would be 1. go back to the original implementation where SPI is completely decoupled from DSPI. 2. is to apply Matt's PR (BUT TEST IT AGAINST ALL BOARDS!). 3. continue to use my stop gap solution to work around the changes that I just wish never happened in the first place.
Brian I trust your judgement fully in this matter; I can make all solutions work one way or another.

majenkotech:
There is no reason not to have a DSPI definition, since the DSPi library is far superior to the SPI library. I myself never actively use the SPI library preferring to use DSPI. My preference would be to use this PR so that there will always be a DSPI entry for those of us that prefer DSPI to SPI.  Iy also means there is no duplication of settings and the possibility of mistakes happening between them.

EmbeddedMan:
Ketih,

Fantastic - thanks for the analysis and context.

I'd like to proceed with your option #2, applying Matt's most recent PR,
and then reverting that part of your PR that deals with the broken boards.
I like the way this latest code can make any DSPI port the default SPI port.

I'll merge Matt's PR, and then you can make the little modifications you
need to, When you get that code PRed, we can use the chipkitbot test build
to test (I Have a pontech board at home to test with).

Sound good?

KeithV:
Sure, 
are you going undue my MX7 changes? Or are you going to have me undue them?

Remember to keep the TaskMgr part of the PR.

EmbeddedMan:
Keith, I'll let you make the correct changes.

PR #170, opened 2016-02-14
initial compile recipe fix

xxxajk:
Fixes the problem where compiler line definitions are missing.
Also is optimized by reuse of common cflags.

Briefly: ARDUINO_ARCH_{build.arch} is not defined for the compile of the .ino file, which causes nasty problems when you want to have a sketch that can compile on multiple platforms, or if the sketch has inline-c/c++ added by includes that depend on it to be defined.

For an example sketch that will fail without this patch see: https://github.com/felis/UHS30/blob/master/libraries/USB_HOST_SHIELD/examples/board_qc/board_qc.ino#L40

The example will fail in the actual libraries as well, since they are all using header-style inlined C and C++.

### DIFF
--- a/pic32/platform.txt
+++ b/pic32/platform.txt
@@ -25,12 +25,14 @@ compiler.c.cmd=pic32-gcc
 # Use a high -Gnum for devices that have less than 64K of data memory
 # For -G1024, objects 1024 bytes or smaller will be accessed by
 # gp-relative addressing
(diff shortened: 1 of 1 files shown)


EmbeddedMan:
I have tested this, and it does not appear to break anything that I can tell. I like the use of compiler.c.flags for compiler.cpp flags. Thanks so much Andrew!

xxxajk:
How about the recipe patch?

xxxajk:
Sorry, wrong list....

PR #171, opened 2016-02-14
Fix SPI so that SPISettings are correctly applied.

xxxajk:
The original storage for the was too small for both brg and con variables.

### DIFF
--- a/pic32/libraries/SPI/SPI.h
+++ b/pic32/libraries/SPI/SPI.h
@@ -85,10 +85,10 @@ public:
         init_AlwaysInline(4000000, MSBFIRST, SPI_MODE0);
     }
 private:
-    void init_MightInline(uint32_t clock, uint8_t bitOrder, uint8_t dataMode) {
+    void init_MightInline(uint32_t clock, uint8_t __attribute__((unused)) bitOrder, uint8_t dataMode) {
         init_AlwaysInline(clock, bitOrder, dataMode);
     }
-    void init_AlwaysInline(uint32_t clock, uint8_t bitOrder, uint8_t dataMode)
+    void init_AlwaysInline(uint32_t clock, uint8_t __attribute__((unused)) bitOrder, uint8_t dataMode)
     __attribute__((__always_inline__)) {
         /* Compute the baud rate divider for this frequency.
         */
@@ -96,29 +96,31 @@ private:
 
         /* That the baud rate value is in the correct range.
         */
-        if (brg == 0xFFFF) {
+        if (brg == 0xFFFFU) {
             /* The user tried to set a frequency that is too high to support.
             ** Set it to the highest supported frequency.
             */
-            brg = 0;
+            brg = 0U;
         }
 
-        if (brg > 0x1FF) {
+        if (brg > 0x1FFU) {
             /* The user tried to set a frequency that is too low to support.
             ** Set it to the lowest supported frequency.
             */
-            brg = 0x1FF;
+            brg = 0x1FFU;
         }
 
-        con = (1 << _SPICON_MSTEN);
+        // By including the needed flag here, we have a speed optimization.
+        con = (1 << _SPICON_MSTEN) | (1 << _SPICON_ON);
 
         switch (dataMode) {
             case SPI_MODE0:
-                con |= ((0 << _SPICON_CKP) | (1 << _SPICON_CKE));
+                con |= (1 << _SPICON_CKE);
+                //con |= ((0 << _SPICON_CKP) | (1 << _SPICON_CKE));
                 break;
 
             case SPI_MODE1:
-                con |= ((0 << _SPICON_CKP) | (0 << _SPICON_CKE));
+                //con |= ((0 << _SPICON_CKP) | (0 << _SPICON_CKE));
                 break;
 
             case SPI_MODE2:
@@ -126,12 +128,13 @@ private:
                 break;
 
             case SPI_MODE3:
-                con |= ((1 << _SPICON_CKP) | (0 << _SPICON_CKE));
+                con |= (1 << _SPICON_CKP);
+                //con |= ((1 << _SPICON_CKP) | (0 << _SPICON_CKE));
                 break;
         }
     }
-    uint8_t brg;
-    uint8_t con;
+    uint16_t brg;
+    uint32_t con;
     friend class SPIClass;
 };
 
@@ -180,10 +183,10 @@ public:
 
         inTransactionFlag = 1;
 #endif
-        pspi->sxCon.clr = (1 << _SPICON_ON);
+        // This causes a nasty spike
+        //pspi->sxCon.clr = (1 << _SPICON_ON);
         pspi->sxBrg.reg = settings.brg;
         pspi->sxCon.reg = settings.con;
-        pspi->sxCon.set = (1 << _SPICON_ON);
     }
 
     // Write to the SPI bus (MOSI pin) and also receive (MISO pin)


EmbeddedMan:
Keith and I have both code reviewed this PR, and it looks good to both of us.

EmbeddedMan:
Andrew - after reviewing the code, I have some questions for you. 

1) What prompted these changes? Were you getting warnings from the compiler?
2) The BRG register only has 5 significant bits. Why is it a good idea to define it in our object as 16 bits and use constants that are 16 bits wide?
3) The CON register looks like it is 32 bits wide. Why define it in our object as 16 bits? Wouldn't we want to represent all 32 bits?

EmbeddedMan:
And one I forgot: Adding "U" on the end of  a a constant makes it an "unsigned int" type, which is 32 bits wide for us. What is the importance of doing this if the BRG register is only 5 bits wide?

xxxajk:
Because when you shift bits in the smaller types, the upper parts, the
values are not preserved. What happens here is the hardware gets trash sent
to it, and spi goes into a frame mode. I tested all of this on my 'scope.
The Idea for the transaction code it to block other isr, and preserve
state. What is still missing is restoration.
The other thing that this fixes is a spike on the SCLK line, which happens
when the SPI is stopped. It should have never been turned off.

xxxajk:
In order to explain the effects of everything, does any of you have a
'scope handy?
I can provide a very simple test that shows the problem, but you'll need to
see the effects on a scope.

A few other things to answer in a more pointed way the actual concerns.
U tells it unsigned, yes, but do not assume it becomes 32bit, because it
does not. It will remain 16bit. The only time it will become 32bit is when
you use UL, or fill in the leading zeros. Yes this also actually solves a
problem, because without it, the test was always false.

16bit is needed because the BRG is 9 bits. There is no 9 bit type, and
using 16 is just fine. 8 bits wasn't enough.

The 32bit con storage solves a problem where bits get scrambled up, and
also provides an optimization, where SPI master is on. Transactions never
use slave mode, they can't. 8bits was totally never enough here to store
anything of use, and would cause the SPI to go into some kind of strange
framing mode. I didn't bother to read back the register with the old code,
but you can try this for yourself, if you wish. You won't be getting what
you expect, at all.

I hope this explains stuff better. the old transaction code simply didn't
work at all, and it won't affect non-transaction (legacy) code. All new
code is discouraged from using the old deprecated methods, and libraries
should be updated to the new transaction API to allow them to share the SPI
bus properly.






On Thu, Feb 18, 2016 at 7:46 AM, Andrew Kroll <xxxajk@gmail.com> wrote:

> Because when you shift bits in the smaller types, the upper parts, the
> values are not preserved. What happens here is the hardware gets trash sent
> to it, and spi goes into a frame mode. I tested all of this on my 'scope.
> The Idea for the transaction code it to block other isr, and preserve
> state. What is still missing is restoration.
> The other thing that this fixes is a spike on the SCLK line, which happens
> when the SPI is stopped. It should have never been turned off.
> On Feb 17, 2016 1:57 PM, "Brian Schmalz" <notifications@github.com> wrote:
>
>> And one I forgot: Adding "U" on the end of a a constant makes it an
>> "unsigned int" type, which is 32 bits wide for us. What is the importance
>> of doing this if the BRG register is only 5 bits wide?
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/chipKIT32/chipKIT-core/pull/171#issuecomment-185348809>
>> .
>>
>


-- 

pic32/libraries/SPI/SPI.h
```diff
@@ -180,10 +183,10 @@ class SPIClass {
 
         inTransactionFlag = 1;
 #endif
-        pspi->sxCon.clr = (1 << _SPICON_ON);
+        // This causes a nasty spike
+        //pspi->sxCon.clr = (1 << _SPICON_ON);
```
EmbeddedMan: I really like all the rest of these changes. But this one I have a question about: here we are going to be changing the baud rate register (BRG) and the SPI Control register (CON), and it looks like the previous code disabled the SPI port (_SPICON_ON cleared), made the changes, then enabled the SPI port (_SPICON_ON set). 

So is it OK to change it so that you never clear the _SPICON_ON bit, but just jam new values into BRG and CON without turning the SPI port off and back on again? I couldn't find anything in the PIC32 reference guide that says this is a bad thing, but every one of their pseudo-code sections disables the port, makes changes to these registers, then re-enables the port.

Maybe a compromise would be to change the code so that the disable and re-enable of the SPI port is ONLY done if the new BRG or CON values are different from the existing values. Then, I would assume, it would happen once at the first transaction, and never after that unless you changed something.

xxxajk: Yes, it should be OK to do that.
The SPI is idle, not transmitting any data. I've not encountered any errors.
All that the old code did was turn off master mode-- which made it slave
mode... which you will never be in when you are using transactions. If you
watch on a 'scope, you'll see the bad side-effect of doing this too... You
get a phantom SCLK that is a very short pulse, which is:
1: Not needed
2: could be a source of EMI/RFI
3: could confuse some SPI devices.

Would you like me to test different SPI rates and see if it simply works?
It should be OK to do this since it only happens when there is no data
transferring, and nothing in the datasheet says anything special is needed.




On Fri, Feb 19, 2016 at 12:11 AM, Brian Schmalz <notifications@github.com>
wrote:

> In pic32/libraries/SPI/SPI.h
> <https://github.com/chipKIT32/chipKIT-core/pull/171#discussion_r53423466>:
>
> > @@ -180,10 +183,10 @@ class SPIClass {
> >
> >          inTransactionFlag = 1;
> >  #endif
> > -        pspi->sxCon.clr = (1 << _SPICON_ON);
> > +        // This causes a nasty spike
> > +        //pspi->sxCon.clr = (1 << _SPICON_ON);
>
> I really like all the rest of these changes. But this one I have a
> question about: here we are going to be changing the baud rate register
> (BRG) and the SPI Control register (CON), and it looks like the previous
> code disabled the SPI port (_SPICON_ON cleared), made the changes, then
> enabled the SPI port (_SPICON_ON set).
>
> So is it OK to change it so that you never clear the _SPICON_ON bit, but
> just jam new values into BRG and CON without turning the SPI port off and
> back on again? I couldn't find anything in the PIC32 reference guide that
> says this is a bad thing, but every one of their pseudo-code sections
> disables the port, makes changes to these registers, then re-enables the
> port.
>
> Maybe a compromise would be to change the code so that the disable and
> re-enable of the SPI port is ONLY done if the new BRG or CON values are
> different from the existing values. Then, I would assume, it would happen
> once at the first transaction, and never after that unless you changed
> something.
>
> —
> <https://github.com/chipKIT32/chipKIT-core/pull/171/files#r53423466>.
>



-- 

xxxajk: p.s. I'll be available for a few more hours today if you need anything
right away. I'm done with my pick-n-place machine and reflowing boards for
the evening...

On Fri, Feb 19, 2016 at 1:31 AM, Andrew Kroll <xxxajk@gmail.com> wrote:

> Yes, it should be OK to do that.
> The SPI is idle, not transmitting any data. I've not encountered any
> errors.
> All that the old code did was turn off master mode-- which made it slave
> mode... which you will never be in when you are using transactions. If you
> watch on a 'scope, you'll see the bad side-effect of doing this too... You
> get a phantom SCLK that is a very short pulse, which is:
> 1: Not needed
> 2: could be a source of EMI/RFI
> 3: could confuse some SPI devices.
>
> Would you like me to test different SPI rates and see if it simply works?
> It should be OK to do this since it only happens when there is no data
> transferring, and nothing in the datasheet says anything special is needed.
>
>
>
>
> On Fri, Feb 19, 2016 at 12:11 AM, Brian Schmalz <notifications@github.com>
> wrote:
>
>> In pic32/libraries/SPI/SPI.h
>> <https://github.com/chipKIT32/chipKIT-core/pull/171#discussion_r53423466>
>> :
>>
>> > @@ -180,10 +183,10 @@ class SPIClass {
>> >
>> >          inTransactionFlag = 1;
>> >  #endif
>> > -        pspi->sxCon.clr = (1 << _SPICON_ON);
>> > +        // This causes a nasty spike
>> > +        //pspi->sxCon.clr = (1 << _SPICON_ON);
>>
>> I really like all the rest of these changes. But this one I have a
>> question about: here we are going to be changing the baud rate register
>> (BRG) and the SPI Control register (CON), and it looks like the previous
>> code disabled the SPI port (_SPICON_ON cleared), made the changes, then
>> enabled the SPI port (_SPICON_ON set).
>>
>> So is it OK to change it so that you never clear the _SPICON_ON bit, but
>> just jam new values into BRG and CON without turning the SPI port off and
>> back on again? I couldn't find anything in the PIC32 reference guide that
>> says this is a bad thing, but every one of their pseudo-code sections
>> disables the port, makes changes to these registers, then re-enables the
>> port.
>>
>> Maybe a compromise would be to change the code so that the disable and
>> re-enable of the SPI port is ONLY done if the new BRG or CON values are
>> different from the existing values. Then, I would assume, it would happen
>> once at the first transaction, and never after that unless you changed
>> something.
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/chipKIT32/chipKIT-core/pull/171/files#r53423466>.
>>
>
>
>
> --
> Visit my github for awesome Arduino code @ https://github.com/xxxajk
>
>


-- 

EmbeddedMan: Nah, don't worry about it. I understand the logic of the changes now, and I'm going to merge the PR, then I want to take a little time this weekend to make a few minor tweaks myself and test them out to make this section of code even clearer. Thanks again - taking the time to submit PRs like this is _so_ appreciated!

xxxajk:
You're quite welcome and thank you very much for accepting the patches!

EmbeddedMan:
See PR #183 for for information on this section of code.

PR #175, opened 2016-02-19
Import Lenny and fix wiring_analog table error

majenkotech:
* Imported Lenny configuration from UECIDE
 * Fixed incorrect table reference in wiring_analog
   * It was hard coded to digital_pin_to_analog_PGM[] instead of using the
    digitalPinToAnalog() macro.


EmbeddedMan:
Matt - you'll need to add a section to the one giant boards.txt file as well to have it show up in Arduino IDE.

If you do in fact want it to show up there. :-)

EmbeddedMan:
Also, I think that you need to back out your change to the digitalPinToAnalog() section. I think I made the same change a while back, and Keith found out that this change broke some expected behavior. I'm not sure I completely understand the reasoning without looking into it more, but here's his explanation at the time:

Brian,
 
The reason why this break is because the function digitalPinToAnalog will return pins 1-14 as analog and I “don’t” to happen. By going directly to the array, ONLY the analog pins are returned and I am checking to see if this is a real analog pin (not the virtual 1-14).
 
KeithV

I'm going to see if we can get some more information on what exactly the issue is with using the function rather than the array.

majenkotech:
I think I understand what ne means. There really shouldn't be the need for such hackiness. It means I will have to change the Lenny configuration then since it doesn't have the table, only the macro, as the macro does everything that is needed. If we can get a list of the affected boards from keith I will take a look at the logic and see if there is a better way of doing it that doesn't require direct access to the table.

And yes, I suppose I should have the board showing up in the Ardueenie IDE.

Ignore this PR for now then, I will ditch it and make a new one.

majenkotech:
Actually this looks to only be an issue for MX1/2 boards. The Uno32 doesn't have a digital_pin_to_analog_PGM table, and that is the spec I based the Lenny on.

majenkotech:
I think I understand the logic for this - though why we only do it for PPS based chips I am not entirely clear.  The tables / macros are actually badly named, since they don't really do what they purport to do.  The macro maps between the pin numbers as used in `analogRead()` and the digital pin numbers that should be used for those values (typically re-mapping 0-5 to 14-19), and the table maps the physical pin numbers to the logical analog pin numbers (eg mapping 14-19 to 0-5) to use for selecting which channel is needed from the analog_pin_to_channel_PGM table.  It's all a bit convoluted really.

KeithV:
Hi Brian,

Somehow I missed your last comment on this. Matt is correct in that it is convoluted; but it has to do with Arduino compatibility. When calling analogRead in Arduino you can either put in the analog number 0-5 (chipkit 0-11) or the digital pin number (14-19,25); both should work on analogRead. This does create a problem if there is an overlap, say we have 20 analog pins, it would be 1-19 or 14-33 and someone specified 15 do they mean analog input 15 or digital pin 15 and they would be different. When making hardware we are careful not to provide more than 15 analog pins so we don't hit pin 14, the first overlap. A potential problem when the MZ has 48 potential analog inputs. In the code at accesses the array directly I needed to know if a "digital" pin was an analog pin and I had to bypass analog_pin_to_channels mapping of 0-11 (chipkit) to analog channels. Since digital pins 0-11 were not analog pins, I didn't want to find a channel for those "numbers" which analog_pin_to_channel would return.

EmbeddedMan:
Keith, very good. Thanks much to you and Matt for helping to document in
this PR the details of why this code needs to stay the way it is, for now,
until somebody re-writes it a better way (sounds like there is lots of room
for improvement, but it must be done _very_ carefully so as to not break
any of the 'features' that existing code depends on). I think I'm
understanding the issue much better now.

KeithV:
To answer Matt's question as to why only PPS part. non-PPS parts have a maximum of 16 analog input pins and digital to analog pin control is defined by 1- 16 bit control register indexed by the analog channel (AD1PCFGCLR = (1 << channelNumber)). On PPS part we must find the actual digital pin port as potentially ANY pin can be an analog pin and we must turn on the analog portion before we can make it an analog input. So on PPS part we must find the port register for the pin (not just the AD1PCFG control register as on non-PPS part). That is for PPS

iop = portRegisters(digitalPinToPort(pin));
bit = digitalPinToBitMask(pin);
iop->ansel.set = bit;

Long story short, on non-PPS just need to know analog channel number
AD1PCFGCLR = (1 << channelNumber)

On PPS need to know IO pin Port
iop = portRegisters(digitalPinToPort(pin));
bit = digitalPinToBitMask(pin);
iop->ansel.set = bit;

And we need to know when analogRead is passed a number if that number is an analog channel number, or a digital pin number; thus using the array to tell us if it was a digital pin number or not.

Issue #155, opened 2016-01-10
Proposal: process platform.txt before deployment


ricklon:
Currently, we're keeping track of 2 platform.txt files. Likely 3 because of MPIDE. But the case of chipKIT-core the two we're tracking are pretty similar. It's possible to use platform.txt as a template and substitute the correct settings when built for URL or Manual install. The difference between files is minimal:
```
23c23
< compiler.path={runtime.hardware.path}/pic32/compiler/pic32-tools/bin/
---
> compiler.path={runtime.tools.pic32-tools.path}/bin/
117c117
< tools.pic32prog.path={runtime.hardware.path}/pic32/tools/bin
---
> tools.pic32prog.path={runtime.tools.pic32prog.path}
```
We would pass a variable about whether build is for URL or manual install. Then correct values would be selected. The search and replace would occur putting the correct values in platform.txt.

If the regex replace task can work this would be possible:
https://ant.apache.org/manual/Tasks/replaceregexp.html

ricklon:
Ok. I just did a test and it's possible to do this!  I'm not sure when these replace tools got added to ANT but yay, it works.

```
<target name="test" >
	  <replaceregexp file="pic32/platform.txt"
		  match="runtime"
		  replace="oldtime"
		  byline="true"
		  />
  </target>
```

EmbeddedMan:
That's great Rick! I'm just uploading a complete set of manual builds right
now that will be the v1.1.0 manual builds. I'll link to them in the Wiki
later this afternoon.

*Brian

EmbeddedMan:
Rick, do you have a PR coming that implements this change to build.xml? I'd love to give it a test.

ricklon:
I can do it but it needs to be Thursday or Friday. Just finished a 2 day
conference presenting, and running makerspace booth. I need a day or so to
catch up.

--Rick

On Tue, Jan 12, 2016 at 7:58 PM, Brian Schmalz <notifications@github.com>
wrote:

> Rick, do you have a PR coming that implements this change to build.xml?
> I'd love to give it a test.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/chipKIT32/chipKIT-core/issues/155#issuecomment-171116427>
> .
>



-- 
Co-founder
Fair Use Building and Research (FUBAR) Labs
http://fubarlabs.org

EmbeddedMan:
No worries, and no rush. If I get a chance I might peck at it, and I'll
[...]

PR #177, opened 2016-02-22
Issue #155 fix so one platform.txt is used.

ricklon:
This adds initials commands for the web/url distro.  It parses the deployed platform.txt file and then compresses the core as a normal distribution.


ricklon:
I'm going to close this and submit an update.

Issue #182, opened 2016-02-24
Fix clock rate calculation in SPI.h (beginTransaction())


EmbeddedMan:
Currently the clock rate register value (BRG) is calculated using __PIC32_pbClk. This value is still zero when the BRG register value is calculated in the Settings object initializer, because the init() call in wiring.c has not happened yet when the Settings objects are created.

Refactor the code in SPI.h to calculate the BRG value each time beginTransaction() is called, thus using the right value of __PIC32_pbClk (and, if it changes while the sketch is running, pick up the new value).

PR #183, opened 2016-02-24
Fix for #182 : Fix clock rate calculation in SPI.h (beginTransaction())

EmbeddedMan:
Also added library example showing off various clock rates and SPI modes in a transaction based SPI sketch.

### DIFF
--- a/pic32/libraries/SPI/SPI.cpp
+++ b/pic32/libraries/SPI/SPI.cpp
@@ -117,8 +117,8 @@ void SPIClass::end() {
     restoreInterrupts(sreg);
 }
 
-void SPIClass::usingInterrupt(uint8_t interruptNumber) {
+void SPIClass::usingInterrupt(uint8_t __attribute__((unused)) interruptNumber) {
 }
 
-void SPIClass::notUsingInterrupt(uint8_t interruptNumber) {
+void SPIClass::notUsingInterrupt(uint8_t __attribute__((unused)) interruptNumber) {
 }
--- a/pic32/libraries/SPI/SPI.h
+++ b/pic32/libraries/SPI/SPI.h
@@ -85,16 +85,52 @@ public:
         init_AlwaysInline(4000000, MSBFIRST, SPI_MODE0);
     }
 private:
+    /* Note that on PIC32, SPI always shifts out Most Significant Bit first, so the bitOrder parameter
+     * is not used, but needs to be retained for compatibility with existing sketches/libraries */
     void init_MightInline(uint32_t clock, uint8_t __attribute__((unused)) bitOrder, uint8_t dataMode) {
         init_AlwaysInline(clock, bitOrder, dataMode);
     }
     void init_AlwaysInline(uint32_t clock, uint8_t __attribute__((unused)) bitOrder, uint8_t dataMode)
     __attribute__((__always_inline__)) {
+        
+        DesiredSPIClockFrequency = clock;
+        
+        // By including the needed flag here, we have a speed optimization.
+        // We will always set the Master Enable bit, becuase we are always the SPI master
+        // We will always set the ON bit because we always want the SPI peripheral turned on
+        con = (1 << _SPICON_MSTEN) | (1 << _SPICON_ON);
+
+        switch (dataMode) {
+            case SPI_MODE0:     // CKE = 1, CKP = 0
+                con |= (1 << _SPICON_CKE);
+                break;
+
+            case SPI_MODE1:     // CKE = 0, CKP = 0
+                break;
+
+            case SPI_MODE2:     // CKE = 1, CKP = 1
+                con |= ((1 << _SPICON_CKP) | (1 << _SPICON_CKE));
+                break;
+
+            case SPI_MODE3:     // CKE = 0, CKP = 1
+                con |= (1 << _SPICON_CKP);
+                break;
+        }
+    }
+    /* This function computes the proper value for the BRG register 
+     * (baudrate generator divisor). This computation can't be done
+     * in this object's intializer, because the value __PIC32_pbClk 
+     * isn't guaranteed to be correct until after all obejcts have
+     * been intialized and the init() call in wiring.c has had a
+     * chance to run. Also, this value can (theoretically) be changed
+     * at runtime (dynamically) and so we need to re-compute the proper
+     * BRG value each time we start a tranasaction to be safe. */
+    uint16_t GenerateBRG(void) {
         /* Compute the baud rate divider for this frequency.
         */
-        brg = (uint16_t)((__PIC32_pbClk / (2 * clock)) - 1);
+        brg = (uint16_t)((__PIC32_pbClk / (2 * DesiredSPIClockFrequency)) - 1);
 
-        /* That the baud rate value is in the correct range.
+        /* Check that the baud rate value is in the correct range.
         */
         if (brg == 0xFFFFU) {
             /* The user tried to set a frequency that is too high to support.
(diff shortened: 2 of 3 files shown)


EmbeddedMan:
Andrew, if you could, would you test out this PR for me? I hope you can follow my changes - there weren't many. But I just couldn't get things to work using the SPI.h code as it was.

Also, I re-enabled the clearing of _SPICON_ON in the CON register before changing the BRG register in beginTransaction(). In your PR, you stated that this would cause glitches on the CLK line. I'm attaching some screenshots of my Saleae Logic analyzer, the raw data file (which you can import into the free Saleae software if you want to see all of the detail), and the test sketch I used to generate the logic analyzer trace. I can't find any difference in the trace if I turn off the SPI port each call to beginTransaction or not - I see no CLK glitching at all.

So, before I merge this code in, I want to make sure you take a look at it, and maybe even test it out on your system to see if you see the CLK glitching with this code or not.

I feel more comfortable with turning the peripheral off before making changes to its registers,  but ONLY if that causes no problems.
[SPITransactionExample.zip](https://github.com/chipKIT32/chipKIT-core/files/143945/SPITransactionExample.zip)
spitransactionexample
[SPITransactionExample.zip](https://github.com/chipKIT32/chipKIT-core/files/143946/SPITransactionExample.zip)

xxxajk:
You won't see the glitches on a logic tracer. it sadly does not have the resolution. I'll show you on the 'scope in a few hours. 
I hope I can photograph it, because the spike is pretty short, much shorter than the 100MHz bandwidth you got on there.

EmbeddedMan:
Andrew, I did use the 500MS/s rate just looking at the clock line, and
still saw no glitches.

xxxajk:
Analogue tells the real story.

I got this test from a fellow engineer for his product, which just
starts the transaction
asserts /SS
sends out 0x55 0x55
deasserts /SS
ends transaction
He has a similar, but fancier scope, but.....

Look at the second-to-last division on the bottom, after the last byte, a spike on the SCLK line.
Also the MOSI line drops to LOW... that should not happen.... It should remain high unless there is an actual mode change.

glitch_spi_present

​Here is what it looks like with the proper fix Notice there is no SCLK glitch, and the data line remains high.
glitch_spi_fixed
Let's compare this to AVR, with photo from friend's fancy scope and also includes a nice digital trace as well.
spi
I guess I need to fix the overshoot/tune my 'scope, but anyway, that is what the scope shows, v.s. what digital will show. The change is way too fast for the FPGA to see it, and it might even filter that out. If yours has an analogue mode, I would suggest looking at it again, I also did this test was on an older UNO32, so your chip/die version could be different and unaffected.

xxxajk:
That said/demonstrated, I'll try your patch, and let you know if the evil glitch returns

xxxajk:
Spike is still there, but much smaller... Unfortunately to the hardware connected, it still looks like a clock tick and causes the ic to start sending the next byte, which starts with a zero. This particular device is smart enough to reset when /SS is deasserted, but some home brew stuff made from simple shift registers isn't, and will cause false data in either direction, as few of them assume /SS is a sort of clear register.

xxxajk:
Here are fresh photos showing the problem. the second one is with 10X magnification. It's very VERY short, but enough to be considered harmful. Looks like your digital logic probe is a fail ;-)

spi_patch_glitch_1x
spi_glitch_10x_magnification

xxxajk:
With the so-called worrisome unnecessary line commented out, we're back to normality.
when_in_doubt_comment_it_out
Sorry, the 'scope isn't lying.

EmbeddedMan:
Thanks so much for documenting this strange clock behavior. Could you post
the simplest sketch that shows the problem? I'm not able to reproduce it
here (yet).

And those last sope shots that you took were with the changes I made last
night?

It looks to me, from your scope shot, that the runt clock pulse is at least
600ps long. Would you agree?

EmbeddedMan:
Does this sketch show the clock glitch for you with the latest code I committed last night?

#include <SPI.h>

const int slaveAPin = 2;

SPISettings settingsA(40000000, MSBFIRST, SPI_MODE0); // Fastest SPI clk rate supported for 80MHz parts

void setup() {
  pinMode (slaveAPin, OUTPUT);
  digitalWrite (slaveAPin, HIGH);
  SPI.begin(); 
}

void loop() {  
  SPI.beginTransaction(settingsA);
  digitalWrite (slaveAPin, LOW);
  SPI.transfer(0x55);
  digitalWrite (slaveAPin, HIGH);
  SPI.endTransaction();
}

EmbeddedMan:
(Also, I'm going to change my PR to remove the offending line - I'm not going through all of this because I think I'm right. I'm just really curious to see if I can somehow replicate the glitch you're seeing, and if not, figure out why.)

xxxajk:
First you'll need a USB Host shield (an official one, not a clone) for the test to work. pm me and I'll get you my email address, then you can email me your shipping details and I can send you one.
Details: the test will show the effects on the data comming back from the device. Other devices may act the same.

Sketch-- I'll attach it, but you'll also need UHS30 installed, see here: https://github.com/felis/UHS30
That contains everything needed to operate the shield, and to be able to run this particular test.

[halt_55-test.zip](https://github.com/chipKIT32/chipKIT-core/files/145291/halt_55-test.zip)

Scope setup is probe the sclk line and MISO line, sync on /SS. To run the QC code, you also need a jumper from pin 9 to pin 7-- or reqire /INT to pin 7 if you got the shield that is.

EmbeddedMan:
Andrew - wow! I didn't realize I'd need all of that just to reproduce the problem. Shouldn't the simple sketch I posted cause the clock glitch for you? Do you think setting up the USB Host shield and such is necessary to see the glitch?

I tried used a 250MHz/2.5GS/s scope at work today and (using my example sketch above) could not see the clock glitch, and that scope will see things pulses far shorter than 600ns.

xxxajk:
Which board?

On Wed, Feb 24, 2016 at 6:17 PM, Brian Schmalz <notifications@github.com>
wrote:

> Andrew - wow! I didn't realize I'd need all of that just to reproduce the
> problem. Shouldn't the simple sketch I posted cause the clock glitch for
> you? Do you think setting up the USB Host shield and such is necessary to
> see the glitch?
>
> I tried used a 250MHz/2.5GS/s scope at work today and (using my example
> sketch above) could not see the clock glitch, and that scope will see
> things pulses far shorter than 600ns.
>
> —
> <https://github.com/chipKIT32/chipKIT-core/pull/183#issuecomment-188510032>
> .
>



-- 

EmbeddedMan:
I did my experiments on a MAX32 board (MX795). Maybe I'm not seeing the
glitch because it only happens on certain versions of the SPI peripheral on
certain particular PIC32 parts?

xxxajk:
That could be the problem....
Sometimes even the same part will show certain issues if it is an older DIE version... meaning usually an earlier chip.
The reason I test using what I do use is because it is a real test, with actual hardware, in order to see actual effects. Adding hardware can change many things, including available power to the MCU, current draw on the pins, etc. All can have an effect that you might not notice when nothing is connected.

1042CB3 is the date code and lot code on the PIC32MX320F128H-80I/PT chip on the uno32 I have.
Perhaps disable of clock disables other parts of the SPI module. AVR has a strange quirk with SPI as well, where if you use it without turning on the pin for /SS in the module for output, SPI won't work. :-) So it is entirely possible there is a silicon bug.

xxxajk:
Oh and my email is xxxajk@gmail.com ... if you want a UHS board, let me know, gimmie your snail-mail address and I'll get it out to you.

EmbeddedMan:
Andrew, I sent you an e-mail from my brian@schmalzhaus.com account.
Hopefully you got it.

xxxajk:
Never got it, sadly

xxxajk:
I sent that email address a message....

PR #192, opened 2016-03-05
Added chipKIT modified Bridge library

majenkotech:
The Bridge library is designed for connecting a Linux OpenWRT based module as a host and network access device to a microcontroller board. Originally part of the Arduino Yun this is now available as a separate device from Dragino using their HE module.

This bridge library, which is normally bundled with the Arduino IDE, has been modified to run properly with chipKIT boards - most notably it takes into account our _USE_USB_FOR_SERIAL_ flag to select the correct Hardware Serial interface (i.e., switching from using Serial to Serial0 when appropriate).


EmbeddedMan:
I have no good way to test or code review this PR before I merge it - but that's fine in this case, as it's just a new library, and one that will not affect anything anybody is already going. Sounds like a really fun project Matt! I can't wait to see what you're doing with OpenWRT and chipKIT.

majenkotech:
I have been using it live for some time now with a number of different
chipKIT boards. The only change needed was the USB serial. Everything else
is stock.

PR #193, opened 2016-03-07
Combined fixes: SPI, WInterrupt, wiring

xxxajk:
This patch set incorporates the following enhancements and bugfixes:

1: External interrupts can now work with HIGH and LOW states via emulation.
2. wiring corrected the CHANGE FALLING RISING definitions. 
3: SPI library now is properly atomic.
4: variants uC32 and UNO32, added missing definitions,


pic32/cores/pic32/WInterrupts.c
```diff
@@ -263,12 +265,22 @@ void __attribute__((nomips16,at_vector(_EXTERNAL_0_VECTOR),interrupt(IPL4SRS)))
 void __attribute__((interrupt(),nomips16)) ExtInt0Handler(void)
 #endif
 {
-
+	VEC_ENT_CODE(EXT_INT0,INT0IF,PIN_INT0);
```
majenkotech: Is this the right order for things here? This would give an extra execution of the code after the interrupt pin has returned to high - is that really wanted?

    HIGH - idle
    HIGH - idle
    LOW - trigger (IFS changed to 1)
    LOW - retrigger (IFS remains at 1)
    LOW - retrigger (IFS remains at 1)
    HIGH - retrigger (IFS cleared to 1) <- ISR called even though interrupt pin is now HIGH
    HIGH - idle

I think the desired operation should only call the ISR while the input is low - which means wrapping the call to the user's ISR function in a check of the IO pin as well as the decision to set/clear the IFS flag.

pic32/variants/Uno32/Board_Defs.h
```diff
@@ -155,6 +155,9 @@
 #define	PIN_INT3	8
 #define	PIN_INT4	35
 
+#define NOT_AN_INTERRUPT -1
+#define digitalPinToInterrupt(p) ((p) == PIN_INT0 ? 0 : ((p) == PIN_INT1 ? 1 : ((p) == PIN_INT2 ? 2 : ((p) == PIN_INT3 ? 3 : ((p) == PIN_INT4 ? 4 : NOT_AN_INTERRUPT)))))
```
majenkotech: Is this macro actually used for anything? If it's not it should be removed (to avoid future confusion). If it is wanted for something in the future it will have to be implemented in all boards and also a default added to pins_arduino.h in the core.

pic32/cores/pic32/WInterrupts.c
```diff
@@ -255,6 +256,7 @@ void detachInterrupt(uint8_t interruptNum)
     }
 }
 
+#define VEC_ENT_CODE(I,B,P) int c = intmode[I]; IFS0bits.B=0; if (intFunc[I] != 0) (*intFunc[I])(); if(c < CHANGE && c == digitalRead(P)) { IFS0bits.B=1; }
```
majenkotech: It would be preferable to use direct port access rather than digitalRead for the level check. digitalRead is slow and adds lots of overhead to an interrupt that is supposed to be rapid-response. It would be better to have an array of pin information (pointer to the IO structure and the pin mask) to directly access the registers.

xxxajk: Excellent point, but I just used what I know as to be universal, as to avoid problems and complications. Fix it up :-)

pic32/cores/pic32/WInterrupts.c
xxxajk: Yes, it is the correct order, retrigger (back-to-back) interrupts are expected and wanted for the emulation, exactly as if it was doing so in real hardware.

pic32/variants/Uno32/Board_Defs.h
xxxajk: Yes! it is required for MANY other libraries. It is expected to be available, even if the core does not use it. It needs to also be ported to all other platforms.

pic32/cores/pic32/WInterrupts.c
majenkotech: Retriggering isn't my concern, it's the fact that you only check for the level after running the isr regardless of the level, so you can get one extra triggering after tge level has changed to inactive.

pic32/variants/Uno32/Board_Defs.h
majenkotech: So this is a separate issue in that chipkit core doesn't provide it. This needs dealing with as a separate issue to make sure that it is fully supported by all boards etc.

pic32/cores/pic32/WInterrupts.c
xxxajk: No library is going to loop in an ISR, that's not the correct way, and will fail on other platforms. Libraries don't need to check for the situation, and don't. They depend on the hardware to signal the handler, even if it means another go-round.

pic32/variants/Uno32/Board_Defs.h
xxxajk: I combined it here as it is ISR related, and for simplicity on my part, but yes, it is an issue on other boards that does need to be addressed. As soon as I get the other two Digilent boards, I'll have PR's for you.

pic32/cores/pic32/WInterrupts.c
majenkotech: I don't think you're understandin the point that I am making. Perhaps I am not being clear.

The ISR is triggered while IFS is 1. IFS is cleared if the interrupt level becomes the inactive state. That is all fine. The ISR will repeatedly trigger while the interrupt pin is i  the active state.

However the ISR will still trigger *once* while IFS js 1 and yet the IO pin is in the *inactive* state if that state changes while not inside the ISR ar the time.

pic32/variants/Uno32/Board_Defs.h
majenkotech: I wouod suggest moving this to a separate PR so it  an be built up and tested properly including the default fallback and any special _PGM tables that may be required for othe rboatds. There are many boards that need to have it implemented and tested, and some may need special care with PPS.

pic32/cores/pic32/WInterrupts.c
xxxajk: Note that the situation is only for NON edge triggered IRQs,
It is how all other hardware that is LEVEL based works....
I have situations where an IRQ is cleared while another comes in, and the line will remain in the state that triggered the ISR in the first place. By the time this part of the code is reached, the line should be readable. The case for a race should be automatically handled as well, since we clear on entry.
It will not affect edge ones at all.

pic32/variants/Uno32/Board_Defs.h
xxxajk: I beat it to death for hours on the UHS shield, and you may confirm that with Brian. It is stable.

xxxajk: To be specific, it was tested for 24hours, with a 1millisecond regular SOF pulse, plus other random timed interrupts. Never missed a single IRQ, and worked exactly as intended.

pic32/cores/pic32/WInterrupts.c
majenkotech: I have generated a couple of timing diagrams to better demonstrate my concern.  This is what I understand is how it is intended to work:

isr

Whilst INT is low (for example) the interrupt will keep triggering.

This is my "concern" situation - where INT returns to HIGH *between* successive calls to the ISR. You get one extra call to the ISR with INT set to HIGH instead of set to LOW. At the time the ISR is called the INT pin is in a "not triggered" state. My concern is, should that interrupt actually execute the user's ISR? Since the test for whether IFS should be set or cleared isn't performed until the USER code is executed, and that USER code is executed regardless of the state of INT, is that actually the desired operation? 

isr1

All I am saying is, shouldn't there be a second check around the user code?

* Clear IFS
* Is the INT pin LOW?
  * Yes: execute user ISR
* Is the INT pin LOW?
  * Yes: Set IFS

That would result in this slight difference to the sequence of events:

isr2

I don't doubt that the overall method is sound and reliable, my only concern is this one case of timing and should that fourth user ISR call in the second diagram actually exist, or should it be tested for and only executed if the INT pin is actually low at the time?

xxxajk: Your close in the understanding, but not quite there, I'll make a timing diagram that demonstrates how it actually works.

majenkotech: To put it another way, by adjusting the "long hand" version of your code to add an extra check:

    int c = intmode[EXT_INT0];

    IFS0bits.INT0IF = 0;
    if (c < CHANGE) {
        if ((intFunc[EXT_INT0] != 0) && (c == digitalRead(PIN_INT0))) {
            *intFunc[EXT_INT0];
        }
        if (c == digitalRead(PIN_INT0)) {
            IFS0bits.INT0IF = 1;
        }
    } else {
        if (intFunc[EXT_INT0] != 0) {
            *intFunc[EXT_INT0];
        }
    }

xxxajk: This is what actually happens when the ISR does not satisfy the hardware requesting a level interrupt.
effects

xxxajk: Here it is a little more exaggerated.
effects

majenkotech: You're still not getting what I am on about. I understand all that clearly. I purposely ignored the clear + set of the IFS since it is irrelevant for this. All that matters is the end result at the end of the ISR, not where in the ISR the IFS is set or cleared.

The case I am concerned about is when the INT is ACTIVE at the end of the ISR so that IFS ends up SET but the INT pin becomes INACTIVE shortly afterwards so that the IFS is SET when the INT is INACTIVE. The ISR is then executed because IFS is SET, but INT is INACTIVE not ACTIVE, so that call to the ISR is a SPURIOUS call that should be filtered out.

Your current arrangement relies on the INT pin ALWAYS being set to its INACTIVE state at a point BEFORE the test to see if IFS should be set or not, which might not always be the case.

xxxajk: Basically, clearing IFS early is an optimization. It allows the code to not have to check anything for edge cases. Furthermore it emulates exactly how it would look in hardware that has actual level IRQs. Yes, you want to exit the ISR so that other ISR on the same priority/subpriority get a chance to execute. By doing it this way you also avoid race conditions.

majenkotech: Yes, I know all that already.

You still haven't addressed my concern though.

xxxajk: By the way, there is still one cornercase that is not handled, and I will deal with that in another PR after this is accepted. If we are already LOW or HIGH, we will need to turn on IFS, even if we are using an edge mode...

majenkotech: That could just be a single digitalRead in the attachInterrupt routine.

You should also I suppose ensure that IFS is cleared in detachInterrupt. Shouldn't make any difference since the interrupt is disabled, but best to be sure.

xxxajk: The hardware certainly should be faster than the software. If it is not, then the handler needs to be smart enough to deal with a spurious interrupt.

majenkotech: You cannot make assumptions about what is attached to the interrupt pin nor how "fast" it would be. Your shield happens to be "fast" enough to work with this, but others may not be.

xxxajk: oh, also, if int retriggers by a new edge, the code doesn't worry about that, setting IFS if it is set does nothing. and if it is too late and retriggers in the epilogue, it will still back-to-back.
Hope that explains it.

xxxajk: also, this isn't a change interrupt, it's simply lLOW HIGH RISING FALLING, CHANGE is not supported, but it could be with great pain.

majenkotech: Nope, you're still not seeing my concern.  You're still telling me things I know.

I am deeply intimate with the internal workings of the PIC32 interrupt system.

Can you actually see the case that I am concerned with? Can you describe it to me?

xxxajk: No extra checking is required. hardware in the real world doesn't do it, and you would only slow down response time.

majenkotech: So you are saying it's OK for the user's ISR to be triggered when the INT pin is inactive just because it went inactive a nanosecond after your check for the INT pin level?

majenkotech: By the way, hardware level interrupts don't normally use the IF or equivalent flag. They just trigger if the input is LOW at the moment of checking for interrupt states in hardware. Therefore when it comes time to trigger your interrupt routine if the input isn't in the active state then it shouldn't trigger the ISR.

xxxajk: Depends on the hardware. ARM NVIC works in level mode, and it won't care if the INT hardware clears early. After some slight debouncing,, the flag is set. Once it is set the hardware can't clear it... nor would you want to. If you have say, a rotary encoder, and it flies by too fast and say, misses a notch, that doesn't matter... an event **some event** triggered it. What the handler does to deal with the event is beyond the scope of the actual reason it triggered.

majenkotech: But your handler has the ability to completely miss the fact that it has *stopped* being triggered until the whole thing has been run again - when it needn't be (and shouldn't be) run again.

xxxajk: FWIW, RFI/EMI/EMC is a known hassel when dealing with the sort of thing you described above, and yes, I know that most processors check int states on instruction fetches... If you get toggles that are too fast, then the processor is too slow for what you are attempting to do -- all the more important to just handle it and not worry why... when the handler queries the attached hardware, it will find out the why, and if there is no why, it should be designed to deal with that situation. I've seen this situation before, and the way to solve it is to not use a hammer for a screw, you get a better tool, a faster processor.

xxxajk: If the attached hardware is not clearing the int causing another triggering/spurious int, then the driver needs to add a delay before it returns.... I've never seen that case, but that would be the fix.

majenkotech: Fast pulses are not the issue here. Slow ones are.

majenkotech: Oh, and FWIW, I have never in my life come across a valid case for level interrupts. They are a nasty hack for people that don't know how to deal with interrupts properly.

majenkotech: A delay in an ISR?! Are you crazy?!

xxxajk: Sometimes you have to. UHS30 use sblocking ISRs that can have delays up to 30 seconds... Since they run with interrupts open and low priority, other handlers are happily serviced, and is unaffected. The main code, yep, it sits there, but that doesn't matter since it doesn't matter. I never ever do anything time sensitive in a main loop, I always use a higher priority timer.

xxxajk: Much hardware is level interrupt only, and with some hardware it is the only sane way to deal with it. I find level interrupts to be much more reliable, and less painful. edge stuff, when it is comming in too fast, can give you a really bad day, and the MAX3421E is an excellent example of a chip that can outrun many processors.

xxxajk: Oh and if you want more "Thats crazy!!"/"You can't do that!!" stuff, I use malloc/realloc/free in an ISR in UHS30 too, even printf.... LOL.

majenkotech: Yep, that's crazy all right...

majenkotech: I can see nothing in the MAX3421E that requires level interrupts on the MCU. The "level" referred to in the datasheet is the behaviour of the INT output pin, not the required interrupt type of the MCU. Yes, you can be lazy and use level interrupts on the MCU to repeatedly poll the INT pin, but personally that would be furthest method from my mind.  With the INT pin set to level you don't even need to use an interrupt pin - you can just poll it. You only need to use an interrupt pin when you have the INT set to edge mode.

majenkotech: Personally I would run in edge mode, and trigger on rising edge. When triggered you service one interrupt then clear the associated IF flag followed by the associated interrupt flag in the MAX3421E. If there is still a pending interrupt a new INT pulse will be generated so your ISR can run again - and again until all the interrupts have been serviced.

xxxajk: polling is how the 2.0 and 1.0 version did it, and that has severe issues, thats why I got away from it.

xxxajk: In any case, although beyond the scope of this discussion, level irq can be used anywhere, even if you teach an edge-only system how to do it, such as pic32. It also keeps the code base easier for people to port, and understand. Yes I could have used edge, but actually, on this particular chip, it imposes a penalty--- you can only ACK one IRQ at a time, of the chip gets confused, which means you spin handing one at a time v.s. all at the same time with one shot as it is now... so, level does have its advantages in this case, especially since it does all the work over SPI. If it was PCIX or something, then edge is really the way to go, I agree, but it wouldn't work very well at all when using slow SPI rates. The code as is can work at rates as low as 500KHz SPI... possibly lower, but I haven't tried. Just as long as everything gets done in time to catch the next SOF tick, it will work.

KeithV:
Please be careful when writing code that looks like

IFS0Bits.INT1IF as INT1IF may not always be in IFS"0" on all processors. You need to verify that all processors including the MZ and all future processors will have INT1 on IFS0 and like wise INT2,3,4,5

This is a great way to break other boards if you haven't done your homework.

xxxajk:
Is there a macro that can assist us?

EmbeddedMan:
Keith, this is a great point. What is the 'right' way to handle this
general problem, in your opinion?

xxxajk:
I think we just need to have a macro that defines something like this
instead of having specifics within the code itself in order to make it
generic and easier to maintain.

KeithV:
There are functions, if you look at Gene's code variants implementation he has routines that determine this at runtime and then bases a structure off of the base address for the processor specific sfrs. Ultimately this comes from all of those tables in the variant files. That macro that Matt said, why do we need that? It is there to instruct you to replace it with something relevant for your board. It is a placeholder to help you as a board designer to know that you need to do these kinds of things. You will notice that there are a lot of tables in the variant files that are replacement for macros in the "default" variant headers. Those default header (default64/default100) where never designed to be used by any one board, but only as templates to drive the board designer to provide the correct functionality for the board. I do not know the specific for finding the base address for the interrupt sfrs, but I would go looking for it. If it doesn't exist, then we probably need to add it.... OR, verify that every PIC32 in existence will work with that code. BTW: it is possible that may be the case as the interrupts are low in the IRQ list. However on the MZ there is a 1-1 mapping of vectors to IRQs, so there are a lot more Vectors on the MZ and at the higher IRQs the MZ dramatically diverges from the MX; and code like this would fail.

xxxajk:
All I did was copy basically from what was there already... Look at the
original that it replaces and you will see the same thing basically...

KeithV:
I suspect that code was written or ported back in the day when there were relatively few PIC32 boards and before the variant mechanism existed. Maybe it was an oversight, maybe it actually does work. Okay I give you that you probably didn't make the situation worse, but in my review I would go track this down before checking it into the project. I admit, I never use this code; so I personally would not hit this. Like I said, it is low in the IRQ list... I might work on ALL PIC32s today.

xxxajk:
If you do not have the time I will check it out later tonight and let you
know.

KeithV:
I do not have the time. This is your PR; you need to insure you don't break anyone. However, thinking about it,if this code does exist in the "core" today (I only saw the adds you made in the diffs), you probably are okay, today. The reason is I think the core.a archive is built on every sketch build and we would get a compiler error even if we don't use the code (it won't be linked). If that is truly the case, chances are it works today. One day a new processor might come about and break this. I would still check it out before checking it in. Unfortunately, there is a lot of legacy code that suffers these kinds of problems. SD had a huge problem which we are slowly fixing as does USB. You just exposed yet another area.

KeithV:
Go look in 
...\packages\chipKIT\hardware\pic32\1.1.0\libraries\MRF24G\SPIandINT.h
To see how I handled this problem for the SPI interrupts.

xxxajk:
You won't need that anymore if you use transactions... That takes care of
all of the interrupts Collision problems with SPI

xxxajk:
So I looked around in the headers, and the struct is more-or-less the same and all the ISRs of concern are all within IEC0. I could not find any that fell out of scope. in fact they look like many of the headers could be merged for easier to maintain code. I would not worry about anything breaking until it does.

KeithV:
Thanks Andrew!

 

I appreciate your diligence. 

 

Below is some code that Gene wrote to address the problem in DSPI; although I see some potential for problems in his code as well; for example the MZ really doesn’t have IRQs, only Vectors; but because we have so much code written against vectors I always map Vectors to IRQs in the variants, and code like what Gene wrote will continue to work. But one day these assumptions will break, and we need to keep this in the back of our minds.

 

Thanks

KeithV

 

 

void
DSPI::init(uint8_t irqErr, uint8_t irqRx, uint8_t irqTx, isrFunc isrHandler) {

      /* The interrupt flag and enable control register addresses and
      ** the bit numbers for the flag bits can be computed from the
      ** IRQ numbers for the SPI port. There are 32 IRQ bits in each IFS
      ** and IEC register. For each IFS register, there is a SET, CLR,
      ** and INV register, so the distance (in dwords) from IFS0 to IFS1
      ** is 4. This code assumes that all of the enable control and flag
      ** bits for an SPI controller are in the same IEC and IFS registers.
      */
      pregIec = ((p32_regset *)&IEC0) + (irqErr / 32);      // interrupt enable control register
      pregIfs = ((p32_regset *)&IFS0) + (irqErr / 32);      // interrupt flag register

      bitErr = 1 << (irqErr % 32);  // error interrupt flag/enable bit
      bitRx  = 1 << (irqRx % 32);         // rx interrupt flag/enable bit
      bitTx  = 1 << (irqTx % 32);         // tx interrupt flag/enable bit

    isr = isrHandler;
}

EmbeddedMan:
Hey, I just wanted to say thank you to everybody for putting in so much effort on this PR. I really like the direction the discussion took. I have not had time yet to go over all of the technical details presented here, but I plan to this week. Likely I'll ask that it be split into a few smaller PRs that have separate purposes. More to follow soon-

xxxajk:
Most of it tends to be intertwined. IRQ and SPI transaction code in
particular are interdependent, so proceed with great care if you want
separate chunks to work independently. Normally I would have done several
smaller PRs, but sometimes one big chunk is the way to go.

<rant>
On another note, indirectly related... who is the moron instructor teaching
people that tabs are 4 spaces? I see this horrid practice pop up over the
last 5 to 10 years, and it really makes code hard to follow. Even K&R say
8. 8 space tabs has been since the invention of the typewriter. Research
this yourself if you wish, there is an actual scientific reason for it-- it
makes things easy to read.
Is there any chance that a reformat would be accepted? I'm willing to do
this.
</rant>

EmbeddedMan:
Sure - we'll have no problem getting people to agree on that one Andrew.
:-| At work I insist that people use 2 spaces instead of tabs, although
much of the code we get from vendors uses 3 spaces instead of a tab, so it
all has to be reformatted. Honestly, I can't imagine using 8 spaces per tab
- with just 4 indent levels, you're already 32 spaces in!

xxxajk:
That's why I use 132 column terminal :-)

majenkotech:
I call Jihad on 8 spaces!

majenkotech:
> That's why I use 132 column terminal :-)

So on your terminal 8 spaces looks like 4 spaces on anyone with a sane terminal? Maybe the fault is your choice of terminal size, not other people's choice of tab size?

xxxajk:
Nahhh... besides, if you are indenting that deep, its usually time to
rethink what you are doing.

majenkotech:
Or switch to COBOL... or *heaven forbid*... Python...

xxxajk:
COBOL, hell no! I am not writing Love Letters to a computer!
Python on the other hand I can deal with.

xxxajk:
Been doing more testing, and found out that the default SPI mode has a small, but fixable problem, which only manifests at higher speeds, and it also depends on the board.
I'll be putting in a new PR for this fix, but I think this is the correct thread to post my findings.

Maximum speed allowed on the shield is 26MHz, guaranteed rate. 

```
Board | Without fix | With fix | MCU maximum allowed
------+-------------+----------+-------------------
UNO32 |   20000000  | 25000000 | 25000000
uC32  |   20000000  | 25000000 | 25000000
WF32  |   14000000  | 20000000 | 25000000
WiFIRE| not tested yet. UHS30 needs to steal an unused IRQ yet...
```

The interesting bit here is how poor the WF32 is. I put my 'scope on it and the SCLK waveform is not hitting the peaks and valleys that it should in order for the target to recognize the edges of the clock. Since the other two boards work just fine, it could be the design of the board, or perhaps by chance a buggered MCU with some slow silicon.

Once UHS30 works on the WiFIRE, I'll provide the simple test and the PR that makes stuff work much better. For now, though, I thought I would share my findings.

EmbeddedMan:
Andrew - I've now had a chance to look at the SCLK output from an Uno32, WF32, FubarinoSD and Teensy 3.0. All of the PIC32 based boards look identical - I had to run them at 20MHz in order to get any SPI output (that was why I couldn't see anything last night). The Teensy only runs up to 16MHz with normal SPI, so I captured that too. All boards passed the loopback test. They all look like normal SPI clocks that are very fast and unterminated to me - I guess basically what I would expect to see. If you put a load on the SCLK line it would probably clear up the overshoot pretty well.



teensy30_sclk_16mhz
uno32_sclk_20mhz
wf32_sclk_20mhz
fubarinosd_sclk_20mhz

EmbeddedMan:
And at 24MHz (with a faster SPI library - spi4teensy3) the Teensy's clock looks just as 'bad' as ours does, I think.
teensy30_sclk_24mhz

majenkotech:
Out of interest, what does the proposed "fix" consist of?

xxxajk:
@Brian I may have a weak mcu, happens. We'll see what the hardware says
when I get things in order. Still going over mz family data sheets to make
dyn swi work.

Briefly, it makes the spi work in a way that non-microchip designed devices
expect, specifically timing-wise. I'll post an image in the comments with
that pr that explains it.

KeithV:
Andrew,

 

There should be no difference between the speeds or layout of the WF32 and or uC32/Uno32. However, your sloppy SCLK is probably because the Adrunio UNO boards put SCLK on the same pin, pin 13, as the LED. For pin compatibility we had to do the same thing. We used LED drivers to attempt to keep the SCLK fast, but never the less the 4.7K resistor to the LED transistor’s base will definitely effect the maximum speed of  SCLK. I would expect the signal to be sloppy. I would not expect it to be any different than the uC32/Uno32 as they are identical circuits. 

 

One thing I would check….. Make sure that you are not driving pin13 as an LED/OUTPUT! Make sure Pin13 is an INPUT before turning on the SPI controller. While the SPI controller will override the GPIO, internally it may make a difference if the GPIO output driver is enabled.

 

I suppose there is another possibility. The uC32/Uno32 are MX3s, the WF32 is an MX6. I suppose it is possible that the MX6 SPI SCLK driver is weaker than the MX3. They are different MCUs after all. 

 

KeithV

KeithV:
Andrew,

 

I did a quick check, the MX3 has a rated Ioh of 12ma, and the MX6 has a Ioh < 10ma. The MX6 output drivers are 20% weaker.

 

KeithV

 

From: Keith Vogel [mailto:keithv@krvogel.com] 
Sent: Saturday, March 26, 2016 7:56 PM
To: 'chipKIT32/chipKIT-core' <reply@reply.github.com>; 'chipKIT32/chipKIT-core' <chipKIT-core@noreply.github.com>
Subject: RE: [chipKIT-core] Combined fixes: SPI, WInterrupt, wiring (#193)

 

Andrew,

 

There should be no difference between the speeds or layout of the WF32 and or uC32/Uno32. However, your sloppy SCLK is probably because the Adrunio UNO boards put SCLK on the same pin, pin 13, as the LED. For pin compatibility we had to do the same thing. We used LED drivers to attempt to keep the SCLK fast, but never the less the 4.7K resistor to the LED transistor’s base will definitely effect the maximum speed of  SCLK. I would expect the signal to be sloppy. I would not expect it to be any different than the uC32/Uno32 as they are identical circuits. 

 

One thing I would check….. Make sure that you are not driving pin13 as an LED/OUTPUT! Make sure Pin13 is an INPUT before turning on the SPI controller. While the SPI controller will override the GPIO, internally it may make a difference if the GPIO output driver is enabled.

 

I suppose there is another possibility. The uC32/Uno32 are MX3s, the WF32 is an MX6. I suppose it is possible that the MX6 SPI SCLK driver is weaker than the MX3. They are different MCUs after all. 

 

KeithV

xxxajk:
Nice! Thanks for confirming this.

On Sat, Mar 26, 2016 at 11:21 PM, KeithV <notifications@github.com> wrote:

> Andrew,
>
>
>
> I did a quick check, the MX3 has a rated Ioh of 12ma, and the MX6 has a
> Ioh < 10ma. The MX6 output drivers are 20% weaker.
>
>
>
> KeithV
>
>
>
> From: Keith Vogel [mailto:keithv@krvogel.com]
> Sent: Saturday, March 26, 2016 7:56 PM
> To: 'chipKIT32/chipKIT-core' <reply@reply.github.com>;
> 'chipKIT32/chipKIT-core' <chipKIT-core@noreply.github.com>
> Subject: RE: [chipKIT-core] Combined fixes: SPI, WInterrupt, wiring (#193)
>
>
>
> Andrew,
>
>
>
> There should be no difference between the speeds or layout of the WF32 and
> or uC32/Uno32. However, your sloppy SCLK is probably because the Adrunio
> UNO boards put SCLK on the same pin, pin 13, as the LED. For pin
> compatibility we had to do the same thing. We used LED drivers to attempt
> to keep the SCLK fast, but never the less the 4.7K resistor to the LED
> transistor’s base will definitely effect the maximum speed of SCLK. I would
> expect the signal to be sloppy. I would not expect it to be any different
> than the uC32/Uno32 as they are identical circuits.
>
>
>
> One thing I would check….. Make sure that you are not driving pin13 as an
> LED/OUTPUT! Make sure Pin13 is an INPUT before turning on the SPI
> controller. While the SPI controller will override the GPIO, internally it
> may make a difference if the GPIO output driver is enabled.
>
>
>
> I suppose there is another possibility. The uC32/Uno32 are MX3s, the WF32
> is an MX6. I suppose it is possible that the MX6 SPI SCLK driver is weaker
> than the MX3. They are different MCUs after all.
>
>
>
> KeithV
>
>
>
> From: Andrew Kroll [mailto:notifications@github.com]
> Sent: Saturday, March 26, 2016 2:09 AM
> To: chipKIT32/chipKIT-core <chipKIT-core@noreply.github.com <mailto:
> chipKIT-core@noreply.github.com> >
> Cc: KeithV <KeithV@KRVogel.com <mailto:KeithV@KRVogel.com> >
>
> Subject: Re: [chipKIT-core] Combined fixes: SPI, WInterrupt, wiring (#193)
>
>
>
> Been doing more testing, and found out that the default SPI mode has a
> small, but fixable problem, which only manifests at higher speeds, and it
> also depends on the board.
> I'll be putting in a new PR for this fix, but I think this is the correct
> thread to post my findings.
>
> Maximum speed allowed on the shield is 26MHz, guaranteed rate.
>
> Board | Without fix | With fix | MCU maximum allowed
> ------+-------------+----------+-------------------
> UNO32 | 20000000 | 25000000 | 25000000
> uC32 | 20000000 | 25000000 | 25000000
> WF32 | 14000000 | 20000000 | 25000000
> WiFIRE| not tested yet. UHS30 needs to steal an unused IRQ yet...
>
> The interesting bit here is how poor the WF32 is. I put my 'scope on it
> and the SCLK waveform is not hitting the peaks and valleys that it should
> in order for the target to recognize the edges of the clock. Since the
> other two boards work just fine, it could be the design of the board, or
> perhaps by chance a buggered MCU with some slow silicon.
>
> Once UHS30 works on the WiFIRE, I'll provide the simple test and the PR
> that makes stuff work much better. For now, though, I thought I would share
> my findings.
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub <
> https://github.com/chipKIT32/chipKIT-core/pull/193#issuecomment-201741589>
>
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/chipKIT32/chipKIT-core/pull/193#issuecomment-201980535>
>



-- 

xxxajk:
Got Wifire going, and I'm actually impressed with it.
One caveat is that I can't seem to get 26MHz to be stable on it, although supposedly it can go to 50MHz...
Here is bench marks using UHS 3 with storage, on VFAT. Unfortunately the compiler generates such huge code that it will not fit on UNO32.

```
uC32
Removing '/1MB.bin' file... completed with 0
1MB write timing test  2048 writes, (0), (0),  3214 ms (3 sec)
completed with 0
1MB read timing test 2048 reads, (20),  3585 ms (4 sec)
completed with 0

WF32
Removing '/1MB.bin' file... completed with 0
1MB write timing test  2048 writes, (0), (0),  3216 ms (3 sec)
completed with 0
1MB read timing test 2048 reads, (20),  3688 ms (4 sec)
completed with 0

Wi-Fire
Removing '/1MB.bin' file... completed with 0
1MB write timing test  2048 writes, (0), (0),  2313 ms (2 sec)
completed with 0
1MB read timing test 2048 reads, (20),  2732 ms (3 sec)
completed with 0

```
Not bad considering it is only full speed, has a lot of work to do (managing VFAT), and all the transfers happen over SPI.

I also added some pull-ups, which seems to help a little bit, but this is all with automatic SPI settings, which figures out how fast we can go, best case, without stability. Now that I got those numbers, I can tune the library for all 4 boards, push all the work, and sen a pr for the fast and standardized SPI. 
--TO-DO: SPI isn't yet smart enough to back off as it aught to, which will be in a PR after the speedup fix.

KeithV:
Andrew,

 

I have successfully (but not fully reliably) run the MZ SPI at 100MHz (over clocked the peripheral bus). This of course was a dedicated SPI and did NOT have the SCK also tied to an LED. I suspect the 26MHz limitation is due to that LED. The other SPI ports do not have that LED on it, and can run faster. But in general a dedicated SPI will run at 50MHz on the MZ reliably. Oh… I also ran the MZ processor (seeming successfully) at 400MHz. And by interleaving 4 ADCs together I got an ADC conversion rate of over 24MSPS (8bit resolution) with surprisingly good results; and I did this all on a WiFIRE.

 

FYI: This was my first board that I completely designed myself. While I had great influence on the WF32, Gene Apperson designed all of the other Digilent PIC32 boards. 

 

KeithV

xxxajk:
@KeithV Yeh, I don't doubt it. Also the silly wasted routing to the pins on the side can't be good for signals either (reflections anyone?!?!). I talked to EmbeddedMan (Brian) about various design flaws in the boards, but the important one was easy for me to rework to get a normal SPI port. I'm tempted to rework a jumper for the LED too. FWIW, all the spi pins (13-10) *could* have been regular pins, and did not have to be SPI at all. SPI belongs on the AVR ICSP connector -- what you call SPI connector. AVR's use the same connector for SPI and for SPI debugging, as well as one-wire debug via using the reset line (which I had to change out from /SS) and +5v was missing too (also easy to rework). As a great example of how SPI doesn't have to be located there, take a good look at the Mega2560, DUE, etc... these have secondary SPI else-where, but you don't have to. The standard SPI connection for all shields is on the 2x3 pin header.

KeithV:
If you look in the variants file for the boards you can pick up other SPI ports. For example you can get SPI1 on the following pins of the WiFIRE

 

Pin 5 – SCK1

Pin 7 – SS1

Pin 35 – SD01

Pin 36 -- SDI1

 

But there is not a standard Arduino pin out that I know of for a 2nd SPI; all boards appear to be different.

 

At these frequencies (<50MHz) we are not experiencing reflection problems, and even if we were the moment you plug in a shield who knows what you are getting. The routing on the board is insignificant compared to going through the shield connectors. 

 

I don’t think the 2X3 debugging header was originally considered to be a shield connector. If you notice on the Uno it is labeled ICSP, not SPI; that is why I think they originally routed the SPI to (what they thought) was the shield connector pins 10-13. Clearly the usefulness of the 2x3 as an SPI became readily apparent and on the Due they did label it SPI. But by that time there were many shields built that depended on pins 10-13 for SPI.

 

But the bottom line is, putting an LED on the SPI clock was a pretty naive thing to do.

 

KeithV

xxxajk:
Will be cleaner to implement an LED jumper. Board has enough mods on it already to cope with the incorrect assumptions. As far as ICSP, yes, on AVR it is dual purpose, SPI and ICSP. I think the reason they broke out the pins to the side was to balance the connectors, then later on they decided to keep it on non-avr since most of the shield designers with a clue were just using that connector anyway. I can ask if you'd like to find out exactly why. In any case, shields that use the side pins are uno only and  deprecated. All new designs use the AVR ICSP (SPI) port.

KeithV:
Thanks Andrew,

 

If you can find out if the Arduino Uno header is deprecating the SPI side pins in leu of the 2x3 SPI, that would be nice. I would love to drop that requirement in our SPI spec. However, I am concerned about backwards compatibility. Our shields mostly use the 2x3 header to pick up SPI (except our most popular basic IO shield). Moving forward, should we build an Uno style header it would be nice to know if the SPI on pins 10-13 is still a requirement.

 

BTW: what board are you building?

 

Thanks

KeithV

xxxajk:
If you need to look at the separate bits, just review on my fork individual branches.
Everything works great now.

xxxajk:
Here's a test sketch for the shield.

[_syschk.ino.txt](https://github.com/chipKIT32/chipKIT-core/files/198445/_syschk.ino.txt)

It verifies basic core functionality and figures out maximum SPI clock speed.
This test is not part of the regular library, and only used for testing on new platforms.

EmbeddedMan:
Andrew - I am still planning on working through this PR to give detailed code feedback and get everything resolved. I know you've been waiting for a while on this one.

xxxajk:
Yes, not a problem, since I have been super busy with work anyway.


On Tue, Apr 19, 2016 at 9:16 PM, Brian Schmalz <notifications@github.com>
wrote:

> Andrew - I am still planning on working through this PR to give detailed
> code feedback and get everything resolved. I know you've been waiting for a
> while on this one.
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/chipKIT32/chipKIT-core/pull/193#issuecomment-212193064>
>



-- 

EmbeddedMan:
I'm going to merge this PR in as-is, even though there may be some repercussions of it down the road, as it's been sitting here for a long time, and there's been a lot of good discussion. We want to get the next version of chipKIT-core out ASAP, and this PR will hopefully pave the way for more good work in future versions.

So - when the new version comes out, please try it out, and remember that this stuff has changed, so that if anything looks funny in 1.2.0, we can come back here to refresh on what changed.

xxxajk:
Excellent! Thank you! I'll be able to move forwards with support for PIC32 now.
I've done some testing, and have not found it to affect anything (as I said before).
We're all busy, but that's how it goes some times.
Sometimes a huge patch set is really needed.

PR #195, opened 2016-03-14
Added extra overloading for DSPI transfer function

majenkotech:
This fixes problems with the new compiler unable to work out which transfer function to use when:

* Using a numeric literal
* Using a uint16_t value.

### DIFF
--- a/pic32/libraries/DSPI/DSPI.h
+++ b/pic32/libraries/DSPI/DSPI.h
@@ -175,7 +175,9 @@ void		setTransferSize(uint8_t txsize);
 */
 void		setSelect(uint8_t sel) { digitalWrite(pinSS, sel); };
 uint32_t	transfer(uint32_t bVal);
+uint16_t	transfer(uint16_t bVal) { return((uint16_t) transfer((uint32_t) bVal)); }
 uint8_t	    transfer(uint8_t bVal) { return((uint8_t) transfer((uint32_t) bVal)); }
+int 	    transfer(int bVal) { return((int) transfer((uint32_t) bVal)); }
 void		transfer(uint16_t cbReq, uint8_t * pbSnd, uint8_t * pbRcv);
 void		transfer(uint16_t cbReq, uint8_t * pbSnd);
 void		transfer(uint16_t cbReq, uint8_t bPad, uint8_t * pbRcv);


majenkotech:
The problem was originally highlighted here: http://uecide.org/forum/viewtopic.php?f=9&t=577

EmbeddedMan:
So I'm really struggling with this one. On the one hand, the behavior of the compiler changed, and code that before generated no warnings now generates warnings. Since we can't (I don't think) go back to the old compiler behavior, it would be nice to find a way to change the library code so that no warnings are given.

And when a user has #define MY_SENDBYTE 5 and then does SPI1.transfer(MYSENDBYTE); it would be really nice if the compiler automatically knew that this was a 1-byte transfer and handled it as such.

But the problem comes in that the user might want to be transmitting 4-bytes, where 5 is in the low byte. So unless they use (uint32_t)MY_SENDBYTE vs (int8_t)MY_SENDBYTE, the compiler can't really know which to use (since it will make huge difference in what comes out of the SPI port). 

What this PR does is hides a potentially useful warning that the compiler will emit when the user should be casting an ambiguous value to a uint32_t or int8_t. The philosophy in the PR is that the library should silently allow code that may not be what the user intends. Granted, virtually all the time the code as written by the user is what they want, and this PR removes the warning by the compiler correctly. But in some small cases, a user may really want a 32-bit transfer, and should be casting their data to make this expectation explicit. 

So, accepting the PR means loosening the type checking in DSPI to remove compiler warnings that most of the time are incorrect, but sometimes would really help a user debug why their transfer isn't as long or short as they expected.

But rejecting the PR means that lots of code that used to compile without warnings now generates warnings.

Arrggg. I'm not sure what to do.

JacobChrist:
Users shouldn't be using #defines for literal values, its bad form.  Keep the PR IMHO.

majenkotech:
You're missing one important fact though...

When you use a numeric literal (or a define) the compiler sees it as an int - so it picks the int version of the overloading...

... whihc is 32 bits.

So as long as you have switched to 32 bit mode and you use a literal you get the full 32 bits.

xxxajk:
You got a couple of different options.
1: Write better code using const instead of define This is actually the right way do go about it if you want to not take up much flash e.g. const uint8_t five_8bit = 5; const uint32_t five_32bit = 5;

2: Cast within a define e.g. ((uint8_t)5) ((uint32_t)5) *might* work, but might not.

majenkotech:
The int version of the overloading was added especially to handle numeric literals. There is no issue sending 5 as a 32 bit value with this pr. The compiler knows it is an int. It's the same on other compilers. A numeric literal is, unless you cast or add a suffix, the native signed integer type. Which is int.

xxxajk:
exactly :-)

xxxajk:
Don't forget that <reinterpret-cast> may work even better...

majenkotech:
Oh, and this all has nothing to do with the transfer sizes at all. Even if you use `(uint8_t)5`, if you are in 32 bit mode, you still get 5 transmitted as the low byte of a 4 byte transfer. The transfer size is set by a different function entirely.

The onoy risk is if you try and force a larger than 8 bit value into the 8 bit version of the transfer, say `(uint8_t)3000` but then that is an entirely different kettle of fish and will generate warnings of its own about type overflows.

majenkotech:
We have always had this overloading. This PR just adds more options in order to help the new compiler.

EmbeddedMan:
Ahh, I think I understand it better now. You guys are all so smart.

I want to recap so that I'm clear on this: because we set the transfer size explicitly when initializing the SPI object, it doesn't matter which version of the single parameter version of transfer() gets called (i.e. it doesn't make a difference what the size is for the data passed into that function) because the previously set number of bytes will always get sent out and received for that transfer. 

So the PR can't hide a valid or useful compiler warning, because in this case the compiler warning never adds any helpful information, because the user has already set the transaction size.

Did I get it right this time?

majenkotech:
Yep, you got it - except that the size is not set when constructing but (at any time the user chooses) using the `setTransferSize()` function. They can switch at will for different situations. Most never do though and just stick to 8 bit.

EmbeddedMan:
Ahh, right. I haven't played with this code enough to have a good feel for all of the options. 

After all of this discussion (which was really good, positive, and helpful) it feels to me like this PR is a good idea for chipKIT-core right now.

PR #197, opened 2016-03-18
DTWI scanner

GastonLagaffe2013:
This is a DTWI example sketch to scan all I2C buses for responding devices listing their bus and address. It uses DTWI instead of the old wire.h

Ciao, Mathias


EmbeddedMan:
ok to test

PR #199, opened 2016-04-01
Make handy defines for hardware visible to sketches and libraries

xxxajk:
Also includes parts of the interrupt definition fixes, but it should merge OK anyway, since it did locally.


xxxajk:
Why are all the handy BASE and IRQ define statements disabled as something "Internal"?
Having them unavailable to libraries and sketches makes it a total nightmare for developing anything that wants to use a custom IRQ for any PIC32, causing one to have to copy and paste from all the Board_Defs.h, then ifdef for each board when this is already provided.
Huge mistake.

EmbeddedMan:
Keith, since this PR impacts some of the Digilent boards, could you do a review and look for any issues?

KeithV:
Bunch of things going on here. It might be a slam dunk to expose them, and it might not. I too have wished these were available to the sketch, in fact, I wish internal didn't really exist at all. But, here is the problem... the more #defines and types you define, the more likely you are to collide with a library or code base someone else wrote somewhere else. You can imagine pulling in the Harmony USB code could have define/type collides and now you have a lot of work to do. Again, this comes down to a code design philosophy. Do you keep as little exposed as possible and effect as little of someone else's code as possible; or do you make everything available so everyone can use it and now deal with collisions. In practice I found exposing seems to be the only real ultimate solution and you just do the work to fix collision. I fear what that might look like with USB Harmony code coming in.

Bottom line, my experience from a practical  matter it is probably slightly better to expose these defines to the sketch; however you have to build every example for every board and insure there are no collisions. Make sure the HTTPServer (both WiFi and PHY) is built for every supported board, and the USB stack for every supported board. If that all passes, you are probably okay.

EmbeddedMan:
Keith, your experience and insight here is very valuable. I agree with you
on both counts. I'll do a set of compiles to make sure there are no
collisions, and we'll then do the merge, and if anybody's library or
sketches have an issue, I hope the file an issue and we can help them
understand what's going on.

*Brian

xxxajk:
This is why in the libraries I design, I prefix the identifier with something unique.
Think of it as something like a namespace. Usually i use two to 5 letters.
For example, UHS3.0, all macro identifiers are prefixed with `UHS_` to prevent collisions.
Personal stuff, I use my initials `AJK_`-- which allows me to blame myself.
While it may seem to be a daunting task, perhaps everything PIC32 specific could be prefixed with-- guess what.... `PIC32_` and I could refactor the entire tree in a couple of hours using Netbeans... because it can refactor an entire tree easily. I'm sure other editors can do the same thing as well, but I know Netbeans has never let me down with code refactoring.

EmbeddedMan:
I have tested this PR with a large number of example sketches, including the HTTPServer, and serial USB examples. I didn't check every single board, but I did check WF32, WiFire, Fubarino SD, Fubarino Mini. I checked some other random example sketches too, like SPI, and Wire. I found no difference in compiling vs v1.1.0 chipKIT-core, so I think we're good to go. There may be some cases that I haven't caught, but it would be good to get this into broader testing for other people to catch problems, so I'm going to merge it.

Thanks Andrew!

xxxajk:
Sweet! I didn't find any problems either. Normally I won't do any sort of
PR without at least some minimal testing.


On Thu, Apr 28, 2016 at 11:16 PM, Brian Schmalz <notifications@github.com>
wrote:

> Merged #199 <https://github.com/chipKIT32/chipKIT-core/pull/199>.
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/chipKIT32/chipKIT-core/pull/199#event-645139385>
>



-- 

xxxajk:
FWIW, you might want to kill the `#ifdef` wrapping all together. Since I personally did not have need for any of the other information, I approached it carefully.

PR #203, opened 2016-04-04
Implemented software-based setBitOrder()

majenkotech:
This sets a variable for the desired bit order, and if that is LSBFIRST the transfer function manually reverses the order of the bits to be sent, and then also reverses the order of bits received. All other values of the variable leave the data alone.

This emulates the AVR's ability to send data through SPI least-significant bit first.


EmbeddedMan:
Nice! I was hoping we'd get around to adding this option some day. Thanks Matt-

PR #206, opened 2016-04-13
Copy Change Notification configuration from Uno32 to uC32.

vutung2311:
As the two CPU has same Change Notification configuration so I think we can copy that configuration from Uno32.

### DIFF
--- a/pic32/variants/uC32/Board_Data.c
+++ b/pic32/variants/uC32/Board_Data.c
@@ -300,6 +300,54 @@ const uint8_t analog_pin_to_channel_PGM[] =
 
 };
 
+const uint32_t digital_pin_to_cn_PGM[] = 
+{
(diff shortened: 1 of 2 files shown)


EmbeddedMan:
ok to test

EmbeddedMan:
Matt/Keith - do either of you guys see any reason that this change would cause problems? It looks good to me.

*Brian

KeithV:
True the Uno32 and uC32 should have the same CN. However, Brian I thought I did this already.... did it not make it into the build, or did we lose it somehow?

EmbeddedMan:
Keith,

Good question - I thought it was in there too, but apparently not. I'll
merge it in.

*Brian

EmbeddedMan:
Many thanks vutung2311 for this PR! Keep them coming-

PR #210, opened 2016-04-20
Added pin mappings for Lenny in SD card hard coded pin defs file

majenkotech:
This hard coded stuff is pure evil. It should be removed once and for all, and also hardware SPI implemented in the SD library. Until then, here's the bit-banging pin definitions for Lenny.


EmbeddedMan:
Looks like a good first step Matt. Thanks-

PR #211, opened 2016-04-25
Brought up to date with master and numerous Harmony USB fixes

majenkotech:
Merged changes from master.  Fixed numerous issues with USB descriptors. Added USB device name / manufacturer and serial number support code.


pic32/cores/pic32/Tone.cpp
```diff
@@ -145,6 +145,9 @@ uint8_t port;
 //************************************************************************
 void disableTimer(uint8_t _timer)
 {
+    // Added to prevent compiler warning about unused paraemer
```
EmbeddedMan: There's a slightly better way to do this with GCC - add __attribute__ ((unused)) right after the parameter. Not important enough to fix, but interesting.

pic32/cores/pic32/WInterrupts.c
```diff
@@ -56,7 +56,7 @@ void __attribute__((interrupt(),nomips16)) ExtInt2Handler(void);
 void __attribute__((interrupt(),nomips16)) ExtInt3Handler(void);
 void __attribute__((interrupt(),nomips16)) ExtInt4Handler(void);
 
-volatile static voidFuncPtr intFunc[NUM_EXTERNAL_INTERRUPTS];
+static volatile voidFuncPtr intFunc[NUM_EXTERNAL_INTERRUPTS];
```
EmbeddedMan: Is this something we need to fix back in the master branch as well?

EmbeddedMan:
I briefly looked at the changed code, and it looks good, although I'm not a USB expert enough to be able to just glance a the changes and know they are correct. But since this is only impacting the Harmony branch (for now), it's a low risk merge, and looks to add some really nice functionality.

majenkotech:
The change in Tone.cpp was brought over from master. It was made by you in December last year. I would have used an attribute.

majenkotech:
The static volatile vs volatile static is a change that came over from master.

PR #214, opened 2016-05-11
Added PONTECH_UAV100 Variant (formally using Default_64)

JacobChrist:
Old Keith V suggestion finally realized.


EmbeddedMan:
OK to test

PR #215, opened 2016-05-12
Inherit TCPSocket from Stream class

majenkotech:
This just adds the correct inheritance chain and fills in the missing blanks with a couple of wrapper functions. 

This now gives the standard Arduino Stream interface to any TCP sockets created with DEIPcK. That means things like print(), println(), find(), findUntil(), etc all now work.  They can also be passed to any functions that expect a Stream object.


EmbeddedMan:
Awesome. I love it Matt - thanks!

PR #219, opened 2016-05-22
Fixed USBSerial::write packet size in HardwareSerial.cpp and step_delay initialization and usage in Stepper.cpp

squintz:
1) Removed Duplicate Copy and Paste from Flush Comment
2) Updated comment to "we can only transmit a maxium of 63 bytes at a time, break it up into 63 byte packets" from "we can only transmit a maxium of 64 bytes at a time, break it up into 64 byte packets"

The following changes are because the current code would seemingly only transmit 62byte packets. 
3) Changed if (size < kMaxUSBxmitPkt) to if (size <= kMaxUSBxmitPkt)
4) if (packetSize >= kMaxUSBxmitPkt) to if (packetSize > kMaxUSBxmitPkt)


squintz:
I found a mistake in my own code. I'll fix this later and submit a new entry.

PR #221, opened 2016-05-27
Fixed I2C and Virtual Program Button configurations. Updated bootloader file

majenkotech:


EmbeddedMan:
Matt, do you want this merged even though chipKITBOT hasn't built properly yet? Or should we poke it until it does, and then merge?

majenkotech:
I would say merge it regardless.  It looks like CKB isn't going to do anything any time soon, and he's not failed, so we can't kick him.

PR #233, opened 2016-07-06
SoftwareSerial with simplex classes

JacobChrist:
This pull request splits the SoftwareSerial class into four classes to realize simplex operation:

SoftwareSeiralBase : public Stream
SoftwareSerialRx : virtual public SoftwareSerialBase
SoftwareSerialTx : virtual public SoftwareSerialBase
SoftwareSerial : public SoftwareSerialRx, public SoftwareSerialTx

It also adds an example for testing the new classes and checking that they are equivalent.

### DIFF
--- a/pic32/libraries/SoftwareSerial/SoftwareSerial.cpp
+++ b/pic32/libraries/SoftwareSerial/SoftwareSerial.cpp
@@ -989,7 +989,7 @@ void SoftwareSerialRx::end()
  * If this RX pin is on a CN, then read from the RX buffer. If not (it's just a normal
  * GPIO pin) then block and read a byte from the RX pin.
  */ 
-int SoftwareSerial::read(void) { return SoftwareSerialRx::read(); }
+//int SoftwareSerial::read(void) { return SoftwareSerialRx::read(); }
 //int SoftwareSerialTx::read(void) { return 0; }
 int SoftwareSerialRx::read(void)
 {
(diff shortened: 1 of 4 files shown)


EmbeddedMan:
chipkit reset this please

EmbeddedMan:
chipkitbot retest this please

EmbeddedMan:
Thanks so much chipkitbot! You're feeling a lot better today. Good for you.

adamwolf:
ok to test

adamwolf:
retest this please

(This is to chipkitbot, not to any humans.)

EmbeddedMan:
@JacobChrist  - I'm wondering about the status of this pull request. Is this still something you think would be a good idea to merge in?

@majenkotech - any concerns with merging this code in?

I'd like to ether merge this guy in as is, fix it (if it is worth working on, but needs work), or kill it if it's not worth it. Just so that we don't have it hanging out there forever.

JacobChrist:
I think its a great idea, but software serial didn't work so great and I
didn't feel the urge to fix it so I abandoned the effort.  I think the
Simplex classes work as expected.

Jacob Christ
*ProLinear/PONTECH, Inc.*
+1 (909) 652-0670 Phone
http://www.pontech.com

EmbeddedMan:
Jacob, all I see are software serial classes - I don't see a separate simplex class. Do you want to break this PR apart into 'simplex' changes and 'software serial' changes, and only commit the simplex changes?

majenkotech:
The two are intertwined. SoftwareSerial has been split into TX and RX, and then a new SoftwareSerial that uses the two TX and RX classes is implemented.

I've not looked in any detail at this PR yet. SoftwareSerial is something I never use if I can help it (and I can always help it...)

EmbeddedMan:
Right, me too. :-) If they're completely intertwined, is it worth it to try
and pull out the simplex code?

majenkotech:
I would suggest starting a new separate software serial library that is separate (for now) from the core. When it's working nicely it can be pulled into the core as a new library (or a replacement for SoftwareSerial).

This PR can then be dropped, but development of the simplex/duplex system can still go forward.

EmbeddedMan:
OK, executive decision time. I'm going to close this PR - if, in the future, we want a simplex class for hardware and software serial (which would be cool, but probably not at the top of the priority list), it should be a rework of the hardware and software serial libraries. Any software serial work (for its own sake) should be a separate PR.

PR #236, opened 2016-07-13
Fix for issue 213

hormigaAzul:
This solves the problem of non-PPS PIC32s not compiling SoftwareSerial.h. To test it, the SoftwareSerial example code was used. The code was compiled all the boards available on the Platformio system and all the boards compiled except for 3 that failed for other errors.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,4 @@
 dist
 tmp
 
+*.project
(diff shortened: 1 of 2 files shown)


EmbeddedMan:
ok to test

EmbeddedMan:
ok to test

EmbeddedMan:
I tested this on a Fubarino SD, and it not only compiles, but appears to work as well.

PR #243, opened 2016-08-08
Fix bad BRG generation in SPI transactions

majenkotech:
SPI transactions assumed a real clock frequency instead of a clock divisor macro. This now calculates the correct frequency to use for BRG based on the difference between a 16MHz UNO and the PBClock of the PIC32.


adamwolf:
chipkitbot retest this please

PR #253, opened 2016-08-23
Improved PWM

majenkotech:
Two new functions:

* analogWriteFrequency(Hz)
* analogWriteResolution(bits)

Accepts any frequency and any sensible number of bits of resolution (1-16).

Maintains duty cycle of any running PWM when the settings are changed.

### DIFF
--- a/pic32/cores/pic32/wiring.h
+++ b/pic32/cores/pic32/wiring.h
@@ -147,6 +147,8 @@ int		digitalRead(uint8_t pin);
 int		analogRead(uint8_t);
 void	analogReference(uint8_t mode);
 void	analogWrite(uint8_t, int);
+void    analogWriteFrequency(uint32_t);
+void    analogWriteResolution(uint8_t);
 
 
 unsigned long	millis(void);
--- a/pic32/cores/pic32/wiring_analog.c
+++ b/pic32/cores/pic32/wiring_analog.c
@@ -53,12 +53,15 @@
 #include "pins_arduino.h"
 #include "Board_Defs.h"
 
-#define	PWM_TIMER_PERIOD	((__PIC32_pbClk / 256) / 490)
-
 uint32_t	analog_reference = 0;	//default to AVDD, AVSS
 
 uint16_t    pwm_active = 0;			//keeps track of active PWM outputs
 
+static uint32_t _pwmFrequency = 490; // 490Hz default
+static uint32_t _pwmResolution = 8;  // 8 bit resolution
+static uint32_t _pwmPeriod = 637;
+
+static void configurePWM();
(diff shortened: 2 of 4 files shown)


majenkotech:
Test program
````C++
void setup() {
    analogWrite(10, 128);
}

void loop() {
    analogWriteFrequency(500);
    analogWriteResolution(8);
    delay(1000);
    analogWriteResolution(10);
    delay(1000);

    analogWriteFrequency(250);
    analogWriteResolution(8);
    delay(1000);
    analogWriteResolution(10);
    delay(1000);
}
````

majenkotech:
I am noticing a bit of duty cycle drift from that test program - most likely a rounding issue in the recalculation of the OCxRS value.

EmbeddedMan:
THANK YOU THANK YOU for creating a test program for this! Now, where can we
put it such that it is available to folks? Can you make a little example
sketch with some nice comments that could get added to an examples
directory for AnalogWrite?

*Brian

ricklon:
This is great.

majenkotech:
Brian, I would, but there are no examples in the core. Only in libraries. The "core" examples are in the IDE...

EmbeddedMan:
Well, that's pretty crappy. It sure would be nice for core authors to be
able to add examples to the "Examples->Built-In Examples" menu.

Huh.

So, why not create a 'fake' library that only includes an examples folder
for this example? Do you think that would work? At least then we could
point folks to it as a good way to see how these functions are used.

*Brian

majenkotech:
That's exactly what I am experimenting with right now - just got to get the Arduino IDE running with chipkit to test it out.

JacobChrist:
I love the examples library idea.

Jacob

bbbnh:
Does the uC32 support different frequencies on different pins? [This document](http://www.mouser.com/ds/2/268/61143H-42553.pdf) seems to indicate that it might. If so, it would be nice to have analogWriteFrequency(pin, freq) as this would be consistent with Teensy which also has this functionality.

majenkotech:
Hardware wise it would be possible, yes, but the whole of the chipKIT core PWM system would then need to be re-written from scratch. It is all based around one single timer (timer 2) and to split it between T2 and T3 would be a huge undertaking - and which pins would you assign to which timer anyway? Whatever we choose it will be wrong for someone.

So no, while it would be "nice", I can't see it being practical. If you want that kind of functionality you will have to DIY with registers.

bbbnh:
I misunderstood what I was reading re: timers in there, so I assumed incorrectly that there was already more than one timer in use. Not a big deal for me personally, but it was relevant and I was curious.

PR #256, opened 2016-08-25
Fixes for MZ EEPROM emulation

majenkotech:
Add MZ page size definition and move EEPROM memory from kseg0 to kseg1 to remove caching.


JacobChrist:
Its working. I have another idea that I will present in steerage.

PR #264, opened 2016-08-30
Added pin def file for 124 pins

majenkotech:
This is actually the same as for 144 pins by the looks of things. Added as a separate file anyway for both simplicity and future flexibility.

### DIFF
--- a/pic32/cores/pic32/p32_defs.h
+++ b/pic32/cores/pic32/p32_defs.h
@@ -735,6 +735,8 @@ typedef struct {
             #include "pps/pins_mz_64.h"
         #elif (__PIC32_PINS__ == 100)
             #include "pps/pins_mz_100.h"
+        #elif (__PIC32_PINS__ == 124)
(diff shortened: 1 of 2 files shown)


JacobChrist:
I just tried this same change on my side and it still would not compile.  I was getting this message:

C:\Users\Jacob Christ\AppData\Local\Arduino15\packages\chipKIT\hardware\pic32\1.3.1\cores\pic32/p32_defs.h:739:41: fatal error: pps/pins_mz_124.h: No such file or directory

             #include "pps/pins_mz_124.h"

May be windows caching the directory before I created the pps/pins_mz_124.h file.

Jacob

majenkotech:
Maybe you typo'd the filename, or placed it in the wrong place?

JacobChrist:
Just found it.  I called the file "pins_mz_124 .h" (space between 4 and .)

Jacob

majenkotech:
D'oh ;)

PR #266, opened 2016-08-31
Reimplemented the Arduino __FlashStringHelper class 

majenkotech:
...and associated helper functions.

This puts in the standard __FlashStringHelper class and associated member functions for handling it. All it does within those helpers is cast the incoming pointer to a "const char *" and work normally.


EmbeddedMan:
Nice.

Issue #273, opened 2016-10-04
EEPROM library: Do not write FF's in putEeprom (Deeprom.cpp) 


JacobChrist:
It looks like an attempt was started but not finished to avoid writing FF's to the EEPROM table.  The way this would work is that if an entry for a specific address doesn't yet exist in the current table and an attempt to write an 0xFF is made the putEeprom function would return a true.  This would work since the getEeprom funciton already returns true if a specific address doesn't exist in the table.

The advantage of having this feature is that it would result in lower flash writes for a sparsely populated EEPROM resulting in better wear.

PR #274, opened 2016-10-05
Writing 0xFF to emulated eeprom does not increase size of table. 

JacobChrist:
Useful if your eeprom is populated with mostly empty data.

This addresses issue #273

### DIFF
--- a/pic32/libraries/EEPROM/utility/Deeprom.c
+++ b/pic32/libraries/EEPROM/utility/Deeprom.c
@@ -268,18 +268,25 @@ BOOL putEeprom(eeSeg * eeprom, uint32_t address, uint8_t data)
 				   tempSeg.temp.valid = 0;
 				   writeFlashWord((void*)&eeprom[i],tempSeg.data);
 
-				   // If data is 0xFF return
-				//if(data == 0xFF) {
-				//	printf(" and data is 0xFF so done.\n\r");
-				//	return fTrue;
-				//}
-			   }
+				// If data is 0xFF return (case were address had different value)
+				// This could be left out and caught below, but leaving it in speeds
+				// up write since we don't have to find the end of the table to
+				// do the write.
+				if(data == 0xFF) {
+					return fTrue;
+				}
+			}
 		}
 		//If empty eeSeg found save location and break
 		else if(getValid(eeprom[i]) && getTaken(eeprom[i])) {
 			nextAvalible = i;
 			break;
 		}
+	}
+
+	// If data is 0xFF return (case where address has never been written)
+	if(data == 0xFF) {
+		return fTrue;
 	}
 
 	//If I == max size no valid segments exist


EmbeddedMan:
I've tested this change, and I'm struggling a little bit to come up with a test case that really shows it's working well. 

I have a test case, which I explain below, that shows that over a 100,000 random write data, the change saves us from 1 erase cycle (out of 194). I'm looking for some feedback on how to make my tests show the effect of the proposed code change more strongly.

Here is my test case:

I added two lines to flash.c (in the libraries\EEPROM\utility folder):
After  line 91, add
`extern unsigned int __EraseCount__;
`
Then after line 104, add
`	__EraseCount__++;`

This simply increments a counter variable (defined in the sketch below) every time an actual erase is called. The whole point of this change is to reduce the number of erases, so I figured this would be a good way to measure its effectiveness.

Then run the sketch below.  Run it once before making the proposed changes to Deeprom.c, and once after. The before run will give you 194 erases, and the after run will give you 193.

This sketch (where it randomly selects an address and then writes a random byte to it) seemed like a pretty good way of simulating 'worst case' type usage, but maybe this proposed change doesn't really help with 'worst case' usage. If not, then I'd like to find what type of usage it does help with.

```
#include <EEPROM.h>

int i;
int j;

unsigned int __EraseCount__ = 0;

void PrintHex8(uint8_t data) // prints 8-bit data in hex
{
     char tmp[4];
     byte first;
     byte second;
     first = (data >> 4) & 0x0f;
     second = data & 0x0f;
     // base for converting single digit numbers to ASCII is 48
     // base for 10-16 to become lower-case characters a-f is 87
     // note: difference is 39
     tmp[0] = first+48;
     tmp[1] = second+48;
     if (first > 9) tmp[0] += 39;
     if (second > 9) tmp[1] += 39;
     tmp[2] = 0;
     Serial.print(tmp);
}

void PrintEEPROM(void)
{
  int value;
  for (i=0; i < 512;)
  {
    PrintHex8(i>>8);
    PrintHex8(i);
    Serial.print(":");
    for (j=0; j < 32; j++)
    {
      value = EEPROM.read(i);
      PrintHex8(value);
      i++;
    }
    Serial.println();  
  }
}

void setup()
{
  uint32_t addr;
  uint8_t data;
  
  Serial.begin(9600);
  Serial.println();
  Serial.println("Starting EEPROM test");

  // Clear entire eeprom
  EEPROM.clear();

  Serial.println("after EEPROM.clear()");

  Serial.print("EraseCount = ");
  Serial.println(__EraseCount__);

  PrintEEPROM();

  for (i = 0; i < 100000; i++)
  {
    addr = random(512);
    data = random(256);
    EEPROM.write(addr, data);
    if (EEPROM.read(addr) != data)
    {
      Serial.print("Error: Data mismatch. At ");
      Serial.print(addr);
      Serial.print(" wrote ");
      Serial.print(data);
      Serial.print(" read back ");
      Serial.println(EEPROM.read(addr));
    }
  }
  
  Serial.println("After random writing");

  Serial.print("EraseCount = ");
  Serial.println(__EraseCount__);
  
  PrintEEPROM();
}

void loop()
{
}
```

JacobChrist:
The reason I added this code is that I'm working on a project that copies
values from EEPROM to RAM and then back to EEPROM.  Most of the data is
0xFF when read out of EEPROM.  When written back it would be writing a lot
of 0xFF's to memory filling up the table and consuming time.

Jacob

Jacob Christ
*ProLinear/PONTECH, Inc.*
+1 (909) 652-0670 Phone
http://www.pontech.com

JacobChrist:
Thanks for testing it by the way, besides not seeming to save many erase
cycles, would you say that its working?

Jacob

Jacob Christ
*ProLinear/PONTECH, Inc.*
+1 (909) 652-0670 Phone
http://www.pontech.com

majenkotech:
I don't know if there's anything in there for this already (probably not), but how about a little check to see if the target value can be written over the current value and end up correct? That way it could be written to the existing cell without needing to erase first.

If "existing" & "new" == "new" then it can be written.

Something like if you have 0b11010110 and you want to write 0b10000100 it can be done:

    11010110
    10000100 &
    --------
    10000100
    xExExxEx

(x = leave, E = erase)

As long as the change is only turning bits off and not on then there is no need to erase the cell to store the value.

That's more erase cycles saved ...

EmbeddedMan:
Jacob,

Well, it doesn't appear to _hurt_ anything. :-) So that's good.

The thing is that there already is a check to see if the value you are
writing is the value that's already there.  So if you have a lot of 0xFFs
in your EEPROM array, and you try to write a lot more, the code

  //If data is same do nothing
  if(data == getData(eeprom[i])) {
  return fTrue;
  }

is going to terminate the call to putEeprom() whenever the new value is the
same as the old value before your new code gets a chance to run.

Does that make sense to you, or am I missing something basic here?

EmbeddedMan:
I don't know guys - 100,000 random values written to random addresses in a
512 byte address space with just 193 erase cycles seems like the code
that's there is doing an awfully good job of wear-leveling as it is.

*Brian

JacobChrist:
Brian,

The reason I don't want to write 0xFF is that I'm going to fill the table completely every time I save my data.  Every save will require a wipe the table.  Also, I have some timing requirements I'm worried about. If I have to re-write a blank value I will have to write 16K worth of 0xFF's to flash in under 40ms.

I can do most of this outside the core except invalidate when going from a non-0xFF to 0xFF.

Matt,

I found it:

http://ww1.microchip.com/downloads/en/DeviceDoc/60001193B.pdf

From Section 52.10

"Once a word is programmed, it must be erased before it can be programmed again, even if
changing a bit from an erased ‘1’ state to a ‘0’ state."

So right now we are not in compliance with the datasheet.  If we wanted to comply with the datasheet, what we could fix the code by doing this:

When we want to write start at the bottom of the table and go up looking for the address we want to write to.  If we do not find it or if we find it and the value we want to write does not match what is in the table already then we append the table with the new address/value pair.  When we read we start at the bottom move up looking for the first value that has a matching address address since the first value found would be the last one written.  If we don't find the address return 0xFF.

EmbeddedMan:
Jacob,

Why will every save require a wipe of the table? If you're writing 0xFF to the same locations you wrote 0xFF to before, then no erase will be required - no write in fact.

Do you have any way to simulate the type of data patterns you'll be writing to EEPROM with the erase counter I describe above? It would confirm for you how may erase cycles you're really asking for - you could test with and without the fix to see if it made much difference.

JacobChrist:
I my emulated eeprom will be 4096 bytes long which will require all 16384 bytes of the table to save all (mostly 0xff) data. If a single byte changes the table will have to be wiped to save all 4096 values.

JacobChrist:
I can add your counter to verify this as well.

JacobChrist:
I found it...

http://ww1.microchip.com/downloads/en/DeviceDoc/60001193B.pdf

Second paragraph of section 52.10:

Once a word is programmed, it must be erased before it can be programmed
again, even if changing a bit from an erased ‘1’ state to a ‘0’ state.

Jacob


Jacob Christ
*ProLinear/PONTECH, Inc.*
+1 (909) 652-0670 Phone
http://www.pontech.com

EmbeddedMan:
OK, this is good. So maybe you can change the test sketch in such as way as
to simulate this - write 512 0xFFs, then write one value of something else,
then 512 FFs, etc. Maybe that would bring out the reduction in erases more
obviously?

*Brian

JacobChrist:
Brian,

A blank table returns 0xFF for all address, if you write a 0xFF to an
address it will create an entry in the table.  My change just doesn't write
an 0xFF if there is no entry or invalidates an existing table entry if it
does.

Matt,

Testing if a new value can be written without invalidating the current
location is not in there.  If my read of the datasheet is correct, your not
supposed to do this, I took a quick look just now and can't find where I
read that.  The existin code is however already doing this to invalidate an
entry in the table.  So if it is possible its a great idea.  I think this
came up in the past and Digilent did some crazy testing of the EEPROM code
and came to the conclusion that it work and that these things are in the
datasheet to protect Microchip from false claims at edge cases where the
chip is running at max temperature all the time.

Brian,

If we implement Matt's suggesting I bet we can save 25% off those erase
cycles.

Jacob



Jacob Christ
*ProLinear/PONTECH, Inc.*
+1 (909) 652-0670 Phone
http://www.pontech.com

EmbeddedMan:
Jacob, I don't see any reason not to merge in your changes. I'm convinced
that the changes don't hurt anything, so why not, right?

Is there any reason you can think of not to merge it in as is?

*Brian

JacobChrist:
Brian,

I don't see any reason not to merge it in, but I do like the idea of running the 0xFF write test and seeing what the erase count looks like.

I think I'm going to open another issue around the single write issue before erase since I think I have a solution.  The solution may actually make read and write times faster too.

Jacob

EmbeddedMan:
I will really love to see the results from your testing with your use case Jacob, just to see if it's really having the effect you think it is. (I hope it is!)

JacobChrist:
Brian,

I've started testing the EEPROM code with the var you added:

extern unsigned int __EraseCount__; 

As well I exposed another variable that I found that keeps track of size of the table:

extern uint32_t __nextAvailable__; 

There is already a "nextAvalible" in the Deeprom.c, but doesn't seem to do anything.  It looks like its left over from when the code was being developed.  I was thinking that it might be a good thing to add these two variable and expose them permanently.  Minimally have some debug ifdef's that make it easy to turn them on.  Maybe give them names like:

__flash_c_erase_count__
&
__deeprom_c_next_available__

EmbeddedMan:
Jacob, that sounds like a good idea. Have you come up with a test sketch
that actually shows a significant difference in write or erase cycles with
and without the change?

PR #278, opened 2016-10-20
Remove hard coded vector locations from MZ interrupt handlers

majenkotech:
The MZ interrupt handlers were hard coded. This broke things like the MRF WiFi driver for DEIPcK which had their own interrupt handler hard coded. This was in addition to also using setIntVector() to place the ISR in the proper ISR call table - one of which methods was therefore redundant.

This change switches to using the (now considered proper) __USER_ISR function definition to remove the hard-coded-ness of the ISRs.

### DIFF
--- a/pic32/cores/pic32/WInterrupts.c
+++ b/pic32/cores/pic32/WInterrupts.c
@@ -50,11 +50,11 @@
 #include "wiring_private.h"
 
 // forward reference to the ISRs
-void __attribute__((interrupt(),nomips16)) ExtInt0Handler(void);
-void __attribute__((interrupt(),nomips16)) ExtInt1Handler(void);
-void __attribute__((interrupt(),nomips16)) ExtInt2Handler(void);
-void __attribute__((interrupt(),nomips16)) ExtInt3Handler(void);
-void __attribute__((interrupt(),nomips16)) ExtInt4Handler(void);
+void __USER_ISR ExtInt0Handler(void);
+void __USER_ISR ExtInt1Handler(void);
+void __USER_ISR ExtInt2Handler(void);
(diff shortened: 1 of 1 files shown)


EmbeddedMan:
Matt, is this one pretty much done, or do you expect more commits on this PR before it should be merged?

majenkotech:
It's done, but could do with a bit of testing by someone with a different perspective to me.

pic32/cores/pic32/WInterrupts.c
```diff
@@ -279,11 +279,7 @@ void detachInterrupt(uint8_t interruptNum)
 #ifdef PIN_INT0
 //************************************************************************
 // INT0 ISR
-#if defined(__PIC32MZXX__)
-void __attribute__((nomips16,at_vector(_EXTERNAL_0_VECTOR),interrupt(IPL4SRS))) ExtInt0Handler(void)
```
EmbeddedMan: Matt, the __USER_ISR() macro does not capture the at_vector() or interrupt() parts of this __attribute__. Is it really OK to replace these with __USER_ISR?

majenkotech: Yes, because those attributes are redundant. They are used to place the function at a specific location and attach it to a specific interrupt. That is not needed, since the setIntVector function does it for you. With them there you get conflicts. Without them everything works happily.

All that is needed is `nomips16` to force 32-bit code, `interrupt()` to make it use `eret` instead of `jr`, and the `section` to group the interrupts in the right region. The rest is not wanted.

EmbeddedMan: Beauty.

Issue #235, opened 2016-07-13
Missing pins_arduino.h variant defines/constants


bperrybap:
The current pic32 variant files Board_defs.h (which are the equivalent to the Arduino.cc pins_arduino.h files) are missing a few of the Arduino defines & constants.

These are missing: LED_BUILTIN, SDA, SCL.

It is very important that LED_BUILTIN be a #define vs a const integer since existing code uses it for conditional compilation.

It look likes they can easily be created from existing defines.
```
#define LED_BUILTIN PIN_LED1
static const uint8_t SDA =  _DTWIO_SDA_PIN
static const uint8_t SCL = _DTWIO_SCL_PIN
```

It is very unfortunate the the Arduino.cc guys used const values for nearly all of their arduino pin values vs defines as it makes it difficult in cases like this to detect if a constant exists at compile time.

eric-wieser:
> It is very important that LED_BUILTIN be a #define vs a const integer since existing code uses it for conditional compilation.

Another option is to use [the following trick](http://stackoverflow.com/q/636595/102441)
```C++
#define LED_BUILTIN LED_BUILTIN
const uint8_t LED_BUILTIN = PIN_LED1;
```

bperrybap:
@eric-wieser that is a great idea. That gives the best of both worlds. A cpp define as well as a compile time constant.
@JacobChrist , would it make sense to update the core again to use this instead?

And what about the SDA defines?
I'm thinking it would be good to use for those as well.

JacobChrist:
@bperrybap I don't fully understand the consequences of having both a define and a const uint8_t.  Also it would seem like the const uint8_t would need to be defined before the #define.

Agreed, what about the SDA defines?

bperrybap:
The define doesn't have to be before or after the const definition.
The preprocessor will simply replace the string with its value which just happens to be same as the define name. The compiler will then see the value. So the define is effectively a NOP as far as the code is concerned.
For example in the case were both forms of LED_BUILTIN existed (define and const), any code that uses LED_BUILTIN would have the string "LED_BUILTIN" replaced with "LED_BUILTIN"  by the pre-processor and the compiler would then see LED_BUILTIN in the code and use the LED_BUILTIN const definition.

In terms of the SDA & SCL defines & definitions they simply don't exist in the chipkit core.
All the Arduino cores have these symbols so they should be created in the chipkit core for each board variant.

eric-wieser:
[...]

PR #283, opened 2016-10-28
Added LED_BUILTIN to Board_Defs.h

squintz:
The Arduino project has added LED_BUILTIN to their official list of Constants. See https://www.arduino.cc/en/Reference/Constants

They have also updated their examples to use LED_BUILTIN which means it must be defined in variants of Board_Defs.h in order to compile some of their examples. This will likely be an issue in most chipKIT variants.

### DIFF
--- a/pic32/variants/CUI32stem/Board_Defs.h
+++ b/pic32/variants/CUI32stem/Board_Defs.h
@@ -84,6 +84,7 @@
 
 /* Define the pin numbers for the LEDs
 */
+#define LED_BUILTIN 64
 #define	PIN_LED1	64
 
 /* ------------------------------------------------------------ */
--- a/pic32/variants/Cerebot_32MX4/Board_Defs.h
+++ b/pic32/variants/Cerebot_32MX4/Board_Defs.h
@@ -91,6 +91,7 @@
 
 /* Define the pin numbers for the LEDs
 */
+#define LED_BUILTIN 60
 #define	PIN_LED1	60
 #define	PIN_LED2	61
 #define	PIN_LED3	62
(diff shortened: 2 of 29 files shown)


EmbeddedMan:
Unfortunately this needs to be done on a per-board basis, as many boards do not have pin 13 connected to an LED. (Thus the reason Arduino went to a constant rather than using "13", because many of their boards don't have it on 13 anymore either.) Each board's variant file will need to get this define.

majenkotech:
Why not

    #define PIN_LEDBUILTIN PIN_LED1

...?

EmbeddedMan:
Ahhh! Yes, that would work well I think.

*Brian

JacobChrist:
This is the same at #235 and #276.  Also, #235 talks about changes to SPI.

Jacob

EmbeddedMan:
squintz, I think Matt's idea of adding just one 
#define LED_BUILTIN PIN_LED1
in one common place (which will work because every board's variant file will already have PIN_LED1 defined appropriately) may be a better way to go than adding this to every variant file.

JacobChrist:
squintz, I think Matt's idea of adding just one 
#define LED_BUILTIN PIN_LED1
in one common place (which will work because every board's variant file will already have PIN_LED1 defined appropriately) may be a better way to go than adding this to every variant file.

majenkotech:
I don't think he can hear you above the noise of the commits.

squintz:
Oops, sorry! I was in the middle of changing it while you guys were replying. I can change it to PIN_LED1

squintz:
So before I go making the changes again... where would the one common place be that #define LED_BUILTIN PIN_LED1 should go? It only took a couple minutes to add it to each board file so it should only take a few minutes to undo what I did.

JacobChrist:
pins_arduino.h?

squintz:
Makes sense to me. I'll close this pull request and pretend it never happened.

PR #285, opened 2016-10-28
Added LED_BUILTIN to pins_arduino.h

squintz:
LED_BUILTIN is missing from all chipKIT board variants. This change adds it to pins_arduino.h which should map the PIN_LED1 from each board variants Board_Defs.h file to LED_BUILTIN while still allowing it to be specifically defined in Board_Defs.h should someone decide that PIN_LED1 doesn't make sense for their particular board.


EmbeddedMan:
Fantastic. Tested on uC32, and worked great.

PR #287, opened 2016-11-10
MZ EFG ADC Code, analogRead component parts, No fire Support

JacobChrist:
This pull request has a few changes:

1. It adds analogRead() support for generic MZ EFG processors.
2. It adds a PONTECH NoFire board type that is basicly WiFire board without the convoluted code that was needed to support multiple processors with a single boards entry.
3. It breaks down the analogRead() function for all processors (except MZ ECG) into three parts that can be called outside of the core to allow for non-blocking reads of an ADC port.  This sample code below shows how to use the analogRead non-blocking code.  Existing analogRead() and this new code cannot live in the same sketch due to use of global variables that keep track of the state of the read between calls to the non-blocking functions.

``` c++
// This is in wiring.c, I shouldn't have to redefine it here
#define CORETIMER_TICKS_PER_MICROSECOND    (F_CPU / 2 / 1000000UL)

//#define ADC_TEST_PIN A12
#define ADC_TEST_PIN 16

void setup() {
  // put your setup code here, to run once:
  Serial.begin(115200);           // start serial for output
  delay(3000);
  Serial.println("chipKIT-core ADC speed test");
  analogReadConversionStart(ADC_TEST_PIN);
}

void loop() {
  // put your main code here, to run repeatedly:
  static uint32_t start_us = readCoreTimer();
  static uint32_t stop_us = readCoreTimer();

  uint32_t value;

#define FAST_ADC
#ifdef FAST_ADC
  if ( analogReadConversionComplete() ) {
    value = analogReadConversion();
    analogReadConversionStart(ADC_TEST_PIN);
  }
#else
  value = analogRead(ADC_TEST_PIN);
#endif

  stop_us = readCoreTimer(); //micros();

  // Display Loop Time
  Serial.print("ADC_TEST_PIN: ");
  Serial.print(value, DEC);
  Serial.println("");
  Serial.print("F_CPU: ");
  Serial.print(F_CPU, DEC);
  Serial.print(" ");
  float us = (float)(stop_us - start_us) / (float)CORETIMER_TICKS_PER_MICROSECOND;
  Serial.print(us, 3);
  Serial.print(" us, ");
  Serial.print(1.0 / us * 1000000.0, 3);
  Serial.print(" Hz");
  Serial.println("");
  delay(1000);

  start_us = readCoreTimer(); //micros();

}
```


EmbeddedMan:
Jacob,

Normally we try to make a single PR only include the code necessary to fix
or implement a single issue as listed in the issue tracker. Could you
please separate out this PR into three separate PRs, each referencing a
single issue, and then tie the PR back to the issue it fixes? (Feel free to
create issues as necessary)

This is important so that the commit history is as clean as possible if we
have to back out the changes for just one of the issues.

Thanks-
*Brian

JacobChrist:
I can do that...  I was got happy that it was working!

EmbeddedMan:
It's very awesome that you got it working Jacob!!!

*Brian

JacobChrist:
Here's how I'll probably do it...

o, One branch for non-blocking analogRead(), pre MZ analogRead()
o. A branch that will follow on from this that adds the MZ analogRead() that is non-blocking()  (since there was never a purely blocking analogRead() for the MZ parts except in Keith's code).
o. A final branch that makes add the NoFire MZ EFG board.
o. There is also some refactoring of PONTECH variant naming.  I'll put this on a fourth branch.

Jacob

JacobChrist:
Closing and moving to individual pull requests for each of these issues.

PR #291, opened 2016-11-14
Fixed Missing DSPI1 Symbols

squintz:
The DSPI examples are currently broke for ChipKit Pi. After looking into this I noticed what appears to be a copy and paste error. Symbols for DSPI0 were defined twice. The seconds set of symbols also had a comment about needing to look up the pin assignments. I took the opportunity to teach myself about PPS and came up with pins 2 & 3 as suitable pins. Someone please double check me on those. 2 is identified in the schematic as D2/INT2 and 3 is D3/INT1/PWM*. 

After making the changes both examples for DSPI now compile without error.

Resource: http://chipkit.net/wp-content/uploads/2013/11/chipKIT-MPIDE-Board-Variant-Mechanism-Rev-A.pdf

### DIFF
--- a/pic32/variants/ChipKIT_Pi/Board_Defs.h
+++ b/pic32/variants/ChipKIT_Pi/Board_Defs.h
@@ -280,7 +280,7 @@ extern const uint8_t	external_int_to_digital_pin_PGM[];
 /*					Serial Port Declarations					*/
 /* ------------------------------------------------------------ */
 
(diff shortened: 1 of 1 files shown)


JacobChrist:
I would be happy to test but I don't have a chipKIT Pi, I think Brian has one.

EmbeddedMan:
I do, and I am happy to test it out and review the changes.

Brian

Issue #292, opened 2016-11-14
Create PONTECH NoFire variant


JacobChrist:
This variant will be a prototype/template for a board that would use the same processor on a Digilent WiFire Rev C board but without the special ADC tricks created to support multiple chips on a single board.  This will allow for testing non board specific MZ EFG analogRead()'s on a WiFire board.  It will also serve as a working prototype for anyone wishing to create a variant based on the PIC32MZ2048EFG100 chip.

EmbeddedMan:
Love this idea Jacob. Excellent.

JacobChrist:
I am not in love with the name, but I didn't want to confuse WiFire users.
It works great, we have been using this def with WiFire boards.

EmbeddedMan:
This is now merged in.

PR #293, opened 2016-11-14
Pontech variants

JacobChrist:
This pull request address issue #292.  It also does some minor refactor of other PONTECH variants so that the folder naming convention is consistent from board to board.


EmbeddedMan:
Jacob, this all looks good to me, although I don't understand enough about our analog code and the PIC32's analog peripheral to know everything is correct.

How much testing with analog reads did you do on non-MZ boards?

JacobChrist:
The MZ code is from Keith I just split it up. Seems to be working for at
least 5 ADC channels we are using. Sine there is really no MZ board and the
WiFire code is untouched I am pretty sure nothing is broken.

Jacob

JacobChrist:
As a matter of fact the the function names still say WiFire. A fact I
realized recently and should have renamed them to be more generic. This is
for another issue though....

On Nov 27, 2016 4:31 PM, "Jacob Christ" <jacob@pontech.com> wrote:

The MZ code is from Keith I just split it up. Seems to be working for at
least 5 ADC channels we are using. Sine there is really no MZ board and the
WiFire code is untouched I am pretty sure nothing is broken.

Jacob

EmbeddedMan:
Jacob,

Understood - that the intent was only to change behavior for MZ parts. But
from my (brief) code review, it looks like non-MZ parts will also have
_some_ limited changes in the overall flow through ReadAnalog. Have you
tested on at least one non-MZ board to make sure all existing analog
functionality remains working?

*Brian

JacobChrist:
I have not. I have tested one channel on at least 3 mx boards. I can do
this check tomorrow.

Jacob

EmbeddedMan:
Thanks Jacob. If you find any problems, file a new issue so we can track
any changes that are needed.

*Brian

PR #299, opened 2016-11-16
Add serial pin definitions for Max32

eric-wieser:
Extracted from the comments next to digital_pin_to_port_PGM in `:/pic32/variants/Max32/Board_Data.c`

There's already precedent for defining these exact symbol names on other boards

### DIFF
--- a/pic32/variants/Max32/Board_Defs.h
+++ b/pic32/variants/Max32/Board_Defs.h
@@ -300,6 +300,10 @@ extern const uint32_t   digital_pin_to_cn_PGM[];
 #define _SER0_IPL_ISR	IPL2SOFT
 #define	_SER0_IPL		2
 #define	_SER0_SPL		0
+#define _SER0_TX_PIN    1
+#define _SER0_RX_PIN    0
+#define _SER0_RTS_PIN   18
+#define _SER0_CTS_PIN   19
 
 /* Serial port 1 uses UART4 (aka UART1B)
 */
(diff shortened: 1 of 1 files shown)


pic32/variants/Max32/Board_Defs.h
```diff
@@ -300,6 +300,10 @@ extern const uint32_t   digital_pin_to_cn_PGM[];
 #define _SER0_IPL_ISR	IPL2SOFT
 #define	_SER0_IPL		2
 #define	_SER0_SPL		0
+#define _SER0_TX_PIN    1
+#define _SER0_RX_PIN    0
+#define _SER0_RTS_PIN   18
```
EmbeddedMan: For the RTS and CTS pins, I don't think that any of our available UART driver code can use these definitions. I don't think we support hardware handshaking, do we?

In any case, I don't think it can hurt to define these, except if there were to be a library or sketch that checked for these pins to be defined, and if so, assumed that hardware handshaking was being used.

majenkotech: We don't support hardware handshaking, no. Or not yet anyway. I am not aware of any library that does or would ever even think of looking for any definitions like that - they're not exactly standard. They can't hurt.

eric-wieser: One thing to be aware of  here is that the `_SER0` flow control pins are not connected to the FTDI chip, so cannot be used at all in USB serial no matter what the software supports

PR #304, opened 2016-11-18
Flagged MX470 as PPS enabled

majenkotech:
The MX470 didn't have the `__PIC32_PPS__` flag set.


majenkotech:
Yeah, it took me about four days of solid coding to come up with this change ;)

PR #321, opened 2016-12-21
Fix warnings in wiring_analog

eric-wieser:
One variable had no type definition, the other was completely unused


JacobChrist:
Eric,

This line looks like it was removed becuase res is a dead variable.
``` c++
uint32_t res = (1 << _pwmResolution);
```

All these changes seem sound to me. I have not tried to test them.

Jacob

eric-wieser:
Correct, `res` is indeed a dead variable

PR #322, opened 2016-12-21
Fix invalid pointer addition warnings

eric-wieser:
Addition to `void*` pointers is a GCC quirk, not a C feature.

Since we're casting to a `uint32_t*` here anyway, and having interrupt vectors that are not aligned would be crazy, we use `uint32_t` pointer addition.

### DIFF
--- a/pic32/cores/pic32/WSystem.c
+++ b/pic32/cores/pic32/WSystem.c
@@ -208,19 +208,19 @@ void initIntVector(void)
 {
     const IMAGE_HEADER_INFO * pImageHeader = getImageHeaderInfoStructure();
     int i = 0;
-    void * pvOrgIntVec = (void *)pImageHeader->pOrgVector0;
+    uint32_t * pvOrgIntVec = (uint32_t *)pImageHeader->pOrgVector0;
 
     for(i=0; i<NUM_INT_VECTOR; i++)
     {
         // If a compiler installed interrupt handler exits, pre-load it with the original handler
         // However, if a class construtor has already loaded a value in the table, don't replace it
         // There is now a default ISR handler so this replacement almost always occurs
-        if(*((uint32_t *) pvOrgIntVec) != 0xFFFFFFFF  && _isr_primary_install[i] == (isrFunc) &_GEN_EXCPT_ADDR)
+        if(*pvOrgIntVec != 0xFFFFFFFF  && _isr_primary_install[i] == (isrFunc) &_GEN_EXCPT_ADDR)
         {
             _isr_primary_install[i] = (isrFunc) pvOrgIntVec;
         }
 
-        pvOrgIntVec += pImageHeader->cbVectorSpacing;
+        pvOrgIntVec += pImageHeader->cbVectorSpacing / sizeof(*pvOrgIntVec);
     }
 
     return;


JacobChrist:
Eric,

I don't quite follow this change:

``` c++
-        pvOrgIntVec += pImageHeader->cbVectorSpacing;
+        pvOrgIntVec += pImageHeader->cbVectorSpacing / sizeof(*pvOrgIntVec);
```

Why was dividing by the the sizeof(*pvOrgIntVec) not needed before but is now needed?

Jacob

eric-wieser:
> Why was dividing by the the sizeof(*pvOrgIntVec) not needed before but is now needed?

Because `int32_t*` addition is different from `char*` pointer addition. The following code always meets its assertions
```
int* p_int = ...; //anything
char* p_c = reinterpret_cast<char*>(p_int);

int* p_int_next = p_int + 1;
char* p_char_next = p_c + sizeof(int)

assert(p_char_next == reinterpret_cast<char*>(p_int_next))
```

assert(reinterpret_ca

JacobChrist:
I see, so a void* acts like a char* and counts a 1 byte during pointer arithmetic.  This definitely was not at my fingertips.  Makes since to me why you did it now.  Thanks.

eric-wieser:
> I see, so a `void*` acts like a `char*`

Therein lies the warning. According to the C++ standard, it does not. GCC allows it to, but with a warning.

EmbeddedMan:
Jacob - are you comfortable that you understand the change enough to merge it?

JacobChrist:
The change looks pretty straight forward. I'm not familiar at all with this code so I wouldn't even know how to test it.  This looks like it only affects MX chips (which is most everything) for MZ chips there is a different function that just returns.

Maybe Matt can chime in.

majenkotech:
When adding `n` to a pointer the actual address it points to `n` times the size of the value the pointer points at.

A `void *` points at a byte, so it increments it by `n` addresses. A `uint32_t *` points to 4 bytes, so it increments it by `n * 4` bytes.

So:

    void *v = 100;
    v + 4 = 104.

    uint32_t *v = 100;
    v + 4 = 116 (v + 4*4).

    uint32_t *v = 100;
    v + (4 / sizeof(uint32_t)) = 100 + ((4*4)/4) = 104

So the "divide by the size" undoes the extra inherent multiplication of working with a pointer to a larger size whilst still at the same time enforcing the 4-byte alignment that is desired when working with 32-bit data.

There's a good description with pretty pictures here: http://www2.cs.arizona.edu/classes/cs352/fall03/array.pdf

majenkotech:
All this change seems to be doing is moving exactly where the uint32_t * is - from a cast mid-way through the code to the definition of the variable itself.

JacobChrist:
I agree Matt and all of Eric changes seems sound.  I just didn't realize that a void* is byte sized, Eric explained it nicely.  I don't seen anything reason not to merge, I just don't know what this code is for or how to test.

eric-wieser:
In theory you could push this further up the chain and redefine `cbVectorSpacing` to be expressed in words, rather than chars.

I vaguely remember getting another warning at the cast, when using `char*` instead of `uint32_t*`, which is why I moved where the conversion occurs. However, I now can't seem to reproduce it.

EmbeddedMan:
OK, having Jacob and Matt's review of this is good enough for me. Thanks for the great discussion guys - I certainly learned something.

Issue #329, opened 2017-02-23
error: 'size_t' does not name a type - from Printable.h


EmbeddedMan:
In chipKIT-core v1.4.0, for any board, if you compile the WiFi->ConnectNoEncryption example, you get the following errors:

In file included from C:\Users\Brian Schmalz\AppData\Local\Arduino15\packages\chipKIT\hardware\pic32\1.4.0\cores\pic32/IPAddress.h:24:0,

                 from C:\Program Files (x86)\Arduino\libraries\WiFi\src\utility/wifi_drv.h:25,

                 from C:\Program Files (x86)\Arduino\libraries\WiFi\src\WiFi.cpp:20:

C:\Users\Brian Schmalz\AppData\Local\Arduino15\packages\chipKIT\hardware\pic32\1.4.0\cores\pic32/Printable.h:34:13: error: 'size_t' does not name a type

     virtual size_t printTo(Print& p) const = 0;

             ^

In file included from C:\Program Files (x86)\Arduino\libraries\WiFi\src\utility/wifi_drv.h:25:0,

                 from C:\Program Files (x86)\Arduino\libraries\WiFi\src\WiFi.cpp:20:

C:\Users\Brian Schmalz\AppData\Local\Arduino15\packages\chipKIT\hardware\pic32\1.4.0\cores\pic32/IPAddress.h:62:13: error: 'size_t' does not name a type

     virtual size_t printTo(Print& p) const;

             ^

lstandage:
Looks like adding

#include <string.h>

to the file fixes the problem. I'll try to figure out a test, and get it into 1.4.1.

Issue #103, opened 2015-10-26
Fix all warnings in Blink compile with Warnings:all


EmbeddedMan:
Now that we can tell the Arduino IDE to set different warning levels, it sure would be nice if we had zero warnings when we compiled something simple like the blink sketch.

EmbeddedMan:
This is getting closer. Unfortunately it needs to be checked for every variant, as some of the warnings apply to code in the variant files.

PR #341, opened 2017-04-19
Lrs 1.4.1 20170412

lstandage:
I went through to fix a couple of low-hanging fruit issues.

#329 fixed by adding string.h to the Printable.h
#103 address some warnings by fixing Print::print and Print::println warnings, and doing some BSP work.

I went through and compiled the ASCII table sketch for all boards, and addressing some of the most immediate issues.


EmbeddedMan:
ok to test

Issue #345, opened 2017-05-08
Compiler warning about serialEventRun in main.cpp


EmbeddedMan:
Here's the warning:

```
C:\Users\bschmalz\AppData\Local\Arduino15\packages\chipKIT\hardware\pic32\1.4.1\cores\pic32\main.cpp: In function 'int main()':

C:\Users\bschmalz\AppData\Local\Arduino15\packages\chipKIT\hardware\pic32\1.4.1\cores\pic32\main.cpp:64:27: warning: the address of 'void serialEventRun()' will always evaluate as 'true' [-Waddress]

         if (serialEventRun) serialEventRun();

                           ^

```

JacobChrist:
It seems to me that this warning is coming up because serialEventRun() is always defined.  Since its always defined I'm not sure why the if statement is even needed (I can see why you would do such a thing if there could be a condition where it was not defined, say a board with no serial ports that does not include HardwareSerial.h, but that doesn't seem to be the case).  I'm a little perplexed as to why checking for != NULL suppresses the warning but it could be that the compiler is just not smart enough to see that this will always evaluate to true as well?  Why not just drop the if statement and bring it back if we have have a board that doesn't include HardwareSerial.h?

majenkotech:
I am not sure exactly why it's there either, but the Arduino code has it, so I copied it. Verbatim.

I am sure it can probably be reduced down to just `serialEventRun();` safely.

JacobChrist:
There are probably some lower end Atmel chips that don't have serial ports.  Possibly this reason.

EmbeddedMan:
OK, I will amend my pull request.

JacobChrist:
Is there some syntax to put into the c code to suppress the warning.  Like to say, hey, we know this is always true, don't tell us about the warning?  Just so we can keep parity with Arduino?

EmbeddedMan:
I'm not sure with this compiler if you can do that or not. I guess I'd rather just make our code 'right', even though it's not exactly the same as Arduino's then. We've done that in a number of other places in the core too. Since there are no chipKIT boards without hardware UARTs, it makes sense to just remove the check.

EmbeddedMan:
If we did want to suppress the warning, we could do it like this:

```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Waddress"
        if (serialEventRun) serialEventRun();
#pragma GCC diagnostic pop

```

That compiles without warning.

[...]

PR #346, opened 2017-05-08
Fix for issue #345 : serialEventRun compiler warning in main.cpp

EmbeddedMan:


EmbeddedMan:
@majenkotech Any reason this change would cause problems?

PR #351, opened 2017-05-11
Update EEPROM interface for new Arduino standard

majenkotech:
Also updated the examples and added newer ones from the Arduino library.

Briefly tested on WiFire C, uC32, Lenny and Max32.

Due to the pre-existing avr/eeprom.h file in the core that has been updated with the compat layer stubs since it can't pick up a library level one instead, so this library has to go with this core.


majenkotech:
I'm wondering, in future, if it would be good to allow board-level abstraction of the `eeprom_read_byte` and `eeprom_write_byte` functions - that way a board that has an actual EEPROM chip on it can use that instead of flash memory...

EmbeddedMan:
Oooo, yeah! That's a great idea, and shouldn't add too much overhead.

Not that we have any boards right now that could use it, but it would be
nice to have that waiting in case we do.

*Brian

PR #353, opened 2017-05-23
Implement Print::availableForWrite

eric-wieser:
This was added in arduino/Arduino#5789

Not implemented yet for USB serial, as that is harder.

That shouldn't stop this patch being merged though - without this change, some Arduino code won't even compile.

### DIFF
--- a/pic32/cores/pic32/Print.h
+++ b/pic32/cores/pic32/Print.h
@@ -42,10 +42,10 @@ class Print
     void setWriteError(int err = 1) { write_error = err; }
   public:
     Print() : write_error(0) {}
-  
+
     int getWriteError() { return write_error; }
     void clearWriteError() { setWriteError(0); }
-  
+
     virtual size_t write(uint8_t) = 0;
     size_t write(const char *str) {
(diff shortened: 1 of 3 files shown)


pic32/cores/pic32/Print.h
```diff
@@ -55,7 +55,11 @@ class Print
     size_t write(const char *buffer, size_t size) {
       return write((const uint8_t *)buffer, size);
     }
-    
+
+    // default to zero, meaning "a single write may block"
+    // should be overriden by subclasses with buffering
+    virtual int availableForWrite() { return 0; }
+
```
eric-wieser: This is taken straight from arduino/Arduino

EmbeddedMan:
This seems just fine to me. @majenkotech any issues with this PR?

majenkotech:
Looks good to me. Simple test of the TX buffer full bit, and a virtual default function.

Not sure how we would go about implementing this for the USB though - have to dig into the TX implementation.

EmbeddedMan:
Agreed. That's going to be a bit more work.

PR #361, opened 2017-07-18
Support for MikroElektronika's Flip N Click MZ product.

lstandage:
This is the initial board support for the MikroE FlipNClick MZ board.

I have run various sketches to show that the peripherals work. PWM, Analog, SPI, UART, all map correctly and give me the results I expect.

I see activity on the I2C bus, and it looks correct, but I haven't fully been able to validate that it works yet. Test sketches are difficult to work properly. However, the assignments look correct, so I would appreciate a double-check.


EmbeddedMan:
Looks like none of these changes will impact anything we already have in the core, so I'm happy to merge them in. I have no ability to test that they work - I trust your testing Larry!

PR #362, opened 2017-07-26
Tweaked FlipNClick to support 252MHz. 

lstandage:
Required bumping Flash wait states to 4 when speed above 200 MHz.
Fixed EEPROM message in compilation.


EmbeddedMan:
OK to test

Issue #225, opened 2016-06-08
Wire library missing setClock()


bperrybap:
The IDE bundled Wire library includes the function setClock() to allow setting the i2c bus clock frequency. (added both AVR and SAM in IDE 1.5.7)
It would be nice to add this function to the chipkit core to be compatible.
It looks like it wouldn't be hard but would require adding a function to DTWI to do it as currently the setting of the clock rate is done inline in beginMaster() and it looks like you can't just call that again with the frequency from an API function in Wire.

JacobChrist:
Something like this:

void twi_setFrequency(uint32_t frequency)
{
  ptwi->ixBrg.reg = (F_CPU/frequency/2) - (((F_CPU / 10000000) * 26) / 25) - 2;
}

bperrybap:
I'm not familiar enough with the h/w to know.
But setClock() is a function referenced from the Wire object which is in the  TwoWire class.
i.e. Wire.setClock(frequency);

Are there mutiple i2c busses? Can each of the run at a different clock or they all tied to same clock?

if there are multiple i2c buses each with its own master clock, then clock setting support will need to be pushed into the DTWI class so that each instance can support its own clock.

However it is implemented, the function setClock() needs to be in the TwoWire class so it can be referenced from the Wire object in order to be compatible with the arduino.cc Wire library.

Mejdi2017:
In October 2017, Wire.setClock() seem still not implemented?
Is there any plan to implement it in the near future?

majenkotech:
I was actually looking at implementing this the other day, however the way that Wire just calls DTWI methods, and DTWI sets the clock when it initially starts master mode, and Wire doesn't, it just seemed too hard to get it to work. So I gave up and switched to using DTWI directly instead.

Mejdi2017:
Did it work?

I see that DTWI is another library of Arduino15\packages\chipKIT\hardware\pic32\1.4.3\libraries\
I see that on line 147 of DTWI.h there is the lines:

    typedef enum
    {
        FQ0Hz       = 0,
        FQ100KHz    = 100000,
        FQ400KHz    = 400000,
        FQ1MHz      = 1000000
    } I2C_FREQ;

I guess I need to modify the library here...
Pardon my newbeness but could you walk me through it? ^^

bperrybap:
I don' think it is that bad.
I think Jacob's solution is close.
[...]

PR #370, opened 2017-11-17
Added Wire.setClock() & Wire.getClock() to Wire and 10k and 50k clock…

bperrybap:
Here is the code for adding Wire.setClock() and Wire.getClock()
I also added a 10k and 50k clock rates to DTWI  for slower clock support which is needed for certain soft radios.
It will appear that there are more changes to Wire.cpp than there actually are since I fixed some incorrect indention and correct some of the bracing style to make the file consistent. 

I really wish that Wire.setClock() would return the actual clock, but the Arduino boys defined it as a void over three years ago when they added it. While the pic version could safely return the value, it would make portability an issue for applications or libraries that might depend on it.
Because of this the code I did behaves the same way as the Arduino AVR WIRE code. (no return from setClock() )


bperrybap:
This PR is from the discussion in issue #225

EmbeddedMan:
ok to test

bperrybap:
Is that a question about testing? Feel free to run it through some tests.
I tested the new Wire example sketch on Arduino 1.8.5 and chipkit core 1.4.3  (with these new updates)
I also tested it with my hd44780 library talking to a PCF8574 at 10k, 50k, 100k and 400k. (1M is too fast for a PCF8574)

EmbeddedMan:
bperrybap - no, it's a signal to our automated build platform. When anyone submits a PR to chipKIT-core, if one of the admins (like Matt or I) send the text "ok to test" on a single line as a response to the PR, the built-bot takes the PR and does a full chipKIT-core build. If you click the "show all checks" link in the 'all checks have passed' comment above, then click on the 'details' link you'll see the build results, and we can then download and try out your code changes without having to build it ourselves.

pic32/libraries/DTWI/DTWI.cpp
```diff
@@ -926,6 +926,118 @@ bool DTWI::stopMaster(void)
...
+ *
+ *    Return Value:
+ *          A I2C_FREQ enum for the given freqHz
+ *
+ *    Description: 
+ *      Converts a frequency in Hz to an I2C_FREQ enum.
+ *
+ *      Unfortunately, due to beginMaster() using a I2C_FREQ enum vs
+ *      an actual frequency, the mapping of frequencies to enums
+ *      is slightly different dependind on picking an enum based on an
```
EmbeddedMan: Spelling : 'depending'

pic32/libraries/DTWI/DTWI.cpp
```diff
@@ -926,6 +926,118 @@ bool DTWI::stopMaster(void)
...
+ *	Then, it must convert the actual frequency back to a I2C_FREQ enum.
+ *	getClock() must round up from the actual frequency to the nearest enum
+ *      value to enusure that the same clock frequency is created from the
+ *      enum when handed to beginMaster()
+ *
+ * ------------------------------------------------------------ */
+DTWI::I2C_FREQ DTWI::getClock(uint32_t *pfreqHz)
+{
+uint32_t freqHz = 0;
+I2C_FREQ I2Cfreq;
```
EmbeddedMan: Indenting on these two lines : should be 2 spaces before the variable declarations so they line up with the if( below.

pic32/libraries/Wire/Wire.cpp
```diff
@@ -334,46 +336,67 @@ uint8_t TwoWire::read(void)
     if(di2c.read(&data, 1) == 1)
     {
         return((uint8_t) data);
-  }
+    }
 
-	return('\0');
+    return('\0');
```
EmbeddedMan: Thank you for fixing all of the gross indenting and braces around here. Looks so much better!

pic32/libraries/DTWI/DTWI.cpp
bperrybap: That module actually uses 4 spaces of indention.

EmbeddedMan: Ahh, right-o. Then let's go with that-

EmbeddedMan:
@majenkotech Any issues with the code in this PR?

bperrybap:
In terms of any "issues" with the code, something to keep in mind is that
this PR makes Wire.setClock()  return a void just like the Arduino.cc version however it could return a uint32_t of the actual clock and be still be compatible with their current API.
IMO returning the actual clock is how the Arduino boys should have done it but they didn't.
As I mentioned in the OP, it does create a portability issue. However, if returning the actual clock from setClock() will eventually be desired, now would be the time to toss it in, to avoid a backward compatibility issue in the future.
I'm almost leaning a bit towards returning the actual clock as it doesn't break any existing Wire code that uses setClock() and it will avoid a future issue should the Arduino boys ever decide to change setClock() from a void to return the clock. (similar to what they did with the write() function in the Print class)
And if we decide to do this, I might go and lobby them (including with a PR) to add getClock() to their code along with changing setClock() to return the actual clock frequency.

EmbeddedMan:
So that's a really good point - should we do the right thing (return the clock speed from setClock()) and possibly break somebody's existing sketch, or leave it as is to be compatible.

Initially I thought it would be best to be compatible, but I think I've changed my mind. I have a hard time thinking of a sketch out there that would depend on setClock() returning void, so I'm wanting to have it return the actual clock.

bperrybap:
I came the same conclusion as you with respect to API compatibility.

I can fix the code with the tweaks above and modify it to return the actual clock rate.
Then issue a new PR.

Also, I could lobby the Arduino boys to add getClock() and at the same time try to get them modify setClock() to return the actual clock frequency. Maybe even sweeten things by giving them the PRs to do it.

Let me know what you guys want.

EmbeddedMan:
Perfect. Please go ahead with it.

All you have to do is commit your changes to your repo, and update the PR
and your changes should then be seen.

You can try submitting a simple PR to the Arduino version of the wire
library too - and use the fact that chipKIT has a return value as a good
reason for them to change their version. :-)

*Brian

bperrybap:
ok, so here is something I just noticed.
The lines in Wire.cpp pointed out above:
```
    return('\0');
```
is actually indented ok with 4 spaces. However in looking closer, that return is just flat wrong. That return is in the read() function and the read() function is supposed to return an int vs a uint8_t
If there are no characters in the buffer -1 is returned. This code returns a 0 so it looks like a byte with a value of zero is available, when really there was a read error since no data was in the buffer.
Every other core I've seen: Arduino AVR, Arduino sam, Arduino samd, Intel, esp8266, Teensy, Teensy3, Adafruit AVR, all work this way. The pic32 core is non compliant.
This should be fixed but probably should be fixed in a different PR.

EmbeddedMan:
bperrybap - if you can resolve the conflicts in this branch, I can merge it in.

bperrybap:
I guess something changed as there were no conflicts when I did the PR.
Something must have got merged in before this PR that broke my updates.
I'll have to go take a look and see what was done.
I'm sure it is probably something really simple.

bperrybap:
I know what the issue is and what needs to be fixed.
I'm not quite quite sure how to fix it so you can merge it.
Do it need to merge in the current tree back into my local repo, fix it and then push the updates back up?
Will that update the existing PR and fix it so you can merge it in?

EmbeddedMan:
No worries. I'll see if I can resolve it here so we can get this merged in.

bperrybap:
The issue is I did both PRs from the master branch, which I thought would be better in case both were not merged in.
There is a collision, but it is retardedly simple, I would have thought that git would have figured it out since it was only a whitespace issue.
The collision is from one PR fixing some indention, and the other fixing some actual code.
In a small area, they happen to overlap.
Its pretty simple and should be fairly obvious what should be done. 
If you click on the [Resolve conflicts] you want the lower part of the diff conflict.
After looking at the code there is a slight change I'd do to the code in receive() but it isn't necessary.

Here is the code for the two functions:
```
// receive() is pre Arduino 1.x
// receive returns '\0' (null) when there is no data
uint8_t TwoWire::receive(void)
{
    int value;

    value = read();
    if(value < 0)
        value = '\0';

    return((uint8_t)value);
}

// read() is Arduino 1.x
// read returns < 0 when there is no data
// must be called in:
// slave rx event callback
// or after requestFrom(address, numBytes)
int TwoWire::read(void)
{
    byte    data;

    if(di2c.read(&data, 1) == 1)
    {
        return((int) ((uint8_t) data));
    }

    return(-1);
}
```




If want/need me to, I'll do the fix if you can give some slight guidance on the mechanics of getting it pushed back up.

EmbeddedMan:
Awesome. No problem - I'll take care of this shortly and merge it in.
Thanks!

EmbeddedMan:
Oh, and what change would you like in receive()?

*Brian

bperrybap:
The change to receive() doesn't really matter. I'm sure the compiled code is the same.
In the PR in receive() 
```
if(value < 0)
        return('\0');
```
vs this in the code I showed above:
```
if(value < 0)
        value = '\0';
```
The change was mere coding style to have a single return in the function vs multiple.
It does not impact anything. It works either way.

EmbeddedMan:
OK, all taken care of. Merged-

Issue #371, opened 2017-11-19
Wire.read() returns incorrect type and incorrect value when buffer is empty


bperrybap:
I mentioned this in issue #370 
The return value from Wire read() is the wrong type.
read()  should return an int vs a uint8_t
*AND* If there is no data in the buffer, -1 is returned.
So the lower eight bits of the int are the data value if the data value is >= 0
Every other core I've seen: Arduino AVR, Arduino sam, Arduino samd, Intel, esp8266, Teensy, Teensy3, Adafruit AVR, all work this way. 
The pic32 core Wire code returns a zero when there is no data in the buffer so it looks like a byte with a value of zero is available, when really there was a read error since no data was in the buffer.
The pic32 core is non compliant.

This is an easy fix and if there are no objections, I'll generate a PR for it.
--- bill

EmbeddedMan:
Bill, this will be a good fix to chipKIT's wire library. Nice find. We are
always looking for ways to become more compatible with what the other cores
do.

EmbeddedMan:
The PR has been merged so I'm closing the issue.

PR #372, opened 2017-11-19
fixed Wire.read() return type and return value when no data in buffer

bperrybap:
This PR corrects the return type from Wire.read() to be **int** instead of **uint8_t**
The patch is not as simple as it would seem as what is returned from the Arduino pre 1.x function receive() is not the same as as the Arduino 1.x read()
This code preserves receive() behavior and corrects read() type and behavior.

While this code does correct the type and return value from read(), there is a potential that some existing might break if it was depending on current type and behavior of read().
Unfortunately, there is now way to do both:

- correct the return type and return value
- preserve existing return type and existing behavior

There is also a non code changing tweak to DTWI.cpp to remove an unused parameter warning.


bperrybap:
This PR is for issue #371

EmbeddedMan:
ok to test

PR #375, opened 2018-01-26
Fixed CDCACM endpoint halting issue

majenkotech:
I found an issue with CDCACM that would halt endpoints when it shouldn't. This doesn't present a problem on MX chips, but breaks MZ chips nastilly.

Also a mistake with the MZ interrupt handler saw the wrong buffers being sent in a TX interrupt, though I don't think this ever actually manifested itself.


EmbeddedMan:
Nice find Matt!

PR #380, opened 2018-02-13
Partial fix for MZ/DA interrupt bug.

majenkotech:
This is a partial fix for the MZ/DA chips not letting their USB interrupts persist. HS mode is still broken, but this at least allows FS mode to work (which is why it's a *partial* fix).


majenkotech:
I haven't yet confirmed that this fix doesn't break other MZ boards, so don't merge it just yet...

majenkotech:
Ok, I'm happy this doesn't break the WiFire, so that's all good.

PR #386, opened 2018-02-25
Added .romdata section

majenkotech:
Adds a section to the linker script to contain flash data that you don't want to actively include in the output. Ideal for big flash blocks that will be programmed by the rest of the program, such as flash filesystems. Causes the space to be allocated in the memory map but doesn't actively include the data in the output file.

### DIFF
--- a/pic32/cores/pic32/chipKIT-application-COMMON-MZ.ld
+++ b/pic32/cores/pic32/chipKIT-application-COMMON-MZ.ld
@@ -767,6 +767,19 @@ SECTIONS
     *(.rodata1)
     . = ALIGN(4) ;
   } >kseg0_program_mem
+
+  /*
+   * Place constant variables in this area that you don't want included
+   * in the output. Ideal for big arrays for storing a filesystem that
+   * would take forever to program. Also has the effect that reprogramming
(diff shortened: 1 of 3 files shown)


EmbeddedMan:
Ahh, very good. I think the memory recipe lines in platform.txt will need to be updated to include this: 

https://github.com/chipKIT32/chipKIT-core/blob/master/pic32/platform.txt#L101

majenkotech:
I wouldn't know about that kind of thing... that's an arduinism...  It seems likely though.

EmbeddedMan:
OK, I can add it no problem. Can you supply a simple test sketch for me
that uses .romdata so I can make sure it all adds up?

majenkotech:
Sure thing, one mo...

```C++
const uint32_t __attribute__((section(".romdata"))) romdata[8192] = {0};

void setup() {
    Serial.begin(115200);
}

void loop() {
    for (int i = 0; i < 8192; i++) {
        Serial.println(romdata[i], HEX);
    }
}
```

EmbeddedMan:
Beautiful. Thanks. I'll update the PR in a bit.

majenkotech:
I can push the change if you want - may be simpler.

PR #395, opened 2018-03-16
Scripted generation of cpudefs data

majenkotech:
The size of the `cpudefs.h` table is getting unmanageable. As more data is added for peripheral count and presence, and more chips are added, it becomes impossible to work with.

So I have written a perl script that uses the CSV file downloadable from Microchip's 32-bit architecture page to generate the content of `cpudefs.h` for us. It outputs to `stdout`, and should be manually piped to `cpudefs_table.h` which is included by `cpudefs.h`.

The script also adds more flags for:

* `__HAS_DMA__`
* `__NUM_DMA_CHANNELS__`
* `__HAS_USB__`
* `__NUM_USB_MODULES__`
* `__HAS_GPU__`
* `__HAS_ETHERNET__`
* `__FLASH_ROW__`
* `__FLASH_PAGE__`

The script has been run with `32-bit Microcontrollers-2018-03-16.csv` to generate current data.


majenkotech:
Ok... so this PR has killed the old WiFire boards. Not my fault!!!

Microchip seem to not want to admit that they ever made the PIC32MZ EC range of chips. They don't appear at all in their catalog any more. Like they never existed.  So we are going to need to have manual blocks for the old chips that are in use.  I'll build them later today and get another PR in.

EmbeddedMan:
D'oh! How did you ever catch this Matt?

majenkotech:
Simple: I couldn't find my rev c and had to resort to using a rev b...

PR #398, opened 2018-03-21
Introduce HelvePic boards to core

GastonLagaffe2013:
Salut,
I have added the variant files for my HelvePic boards to the variant directory and extended the boards.txt file to add one board to the boards menu. All variations of the board design are entered as menu items to avoid flooding the board list in the IDE menu with entries.
Ciao, Mathias


EmbeddedMan:
OK to test

EmbeddedMan:
I see no problem with these changes.

PR #399, opened 2018-03-23
Expaned keyboard to allow raw usage report operation

majenkotech:
Add four extra methods to the USB Keyboard class:

    pressRaw(k)
    pressRawModifier(k)
    releaseRaw(k)
    releaseRawModifier(k)

They take raw USB HID Usage Code numbers (lots of macros provided) to allow use of extended keys over and above the normal keys on a standard keyboard.


EmbeddedMan:
Cool!!

PR #400, opened 2018-03-24
Reinstated missing E2END

majenkotech:
The `E2END` macro got accidentally left out of the new `cpudefs.h` file.  This breaks EEPROM.

Quite why this is in the core and not either in the linker script (where the EEPROM size is defined) or the EEPROM library is anyone's guess - but it is... or was...

### DIFF
--- a/pic32/cores/pic32/cpudefs.h
+++ b/pic32/cores/pic32/cpudefs.h
@@ -38,6 +38,8 @@
 //************************************************************************
 //*    Microchip pic32 chip names
 
+#define    E2END        0x0fff    //*    4 k of simulated EEPROM
+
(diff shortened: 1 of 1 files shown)


EmbeddedMan:
I think the reason it's not in a linker script is because we wanted it to be a constant value across all chipKIT boards, since we want sketches to be able to depend on it.

Maybe putting it in the EEPROM library would be a better idea though - since any sketch that uses EEPROM will use that EEPROM library, right?  But in actuality, it's also 'used' in the linker script. There are multiple places that need to know this number, and it would be great to only have one place with one number. I'm just not sure I know how to easily do that at this point.

So I think this is good enough .

PR #407, opened 2018-05-19
Added digitalPinToInterrupt generic macro

majenkotech:
A generic macro that compares a pin number to the pins values in PIN_INT0 to PIN_INT4 in turn to find the interrupt number for a pin.  This can be overridden in a Board_Defs.h file if not all 5 PIN_INTx macros are defined for a board, or an unused PIN_INTx can be defined as NOT_INT_PIN to disable it instead.


EmbeddedMan:
Looks good Matt. Thanks.

PR #410, opened 2018-05-25
Added support for HID Keyboard LEDs

majenkotech:
A keyboard device now receives LED reports from the host (Num lock, Caps lock, Scroll lock).

Also has a few extra USB enhancements and optimizations.


EmbeddedMan:
Nice!

PR #414, opened 2018-05-31
implemented the yield() function

bluetiger9:
Hi,

Added the `yield()` function, as some libraries needs it.

I added the declaration  in `WProgram.h`. The implementation is in new `hooks.c` file _(copied from [ArduinoCore-avr](https://github.com/arduino/ArduinoCore-avr/blob/master/cores/arduino/hooks.c)_).

Let me now is some changes are needed.

Thanks,
Attila


majenkotech:
ok to test

EmbeddedMan:
Very nice.

PR #421, opened 2018-06-05
Restored File::readDir to SD library

majenkotech:
Not sure if this was ever actually in the core copy of the library or not, or if I just had it in an old custom UECIDE version, but readDir() is a pretty essential function if you want to examine the contents of a directory.

This exposes (or re-exposes) the internal readDir call through the File class.


EmbeddedMan:
Nice. Can't hurt.

Issue #429, opened 2018-06-26
dtostrf convertig error


pgunzelmann:
example simplified to show only the problem:
```
void setup() {
  // put your setup code here, to run once:
  double d=1.04;
  Serial.begin(9600);
  Serial.println(String(d));
}

void loop() {
  // put your main code here, to run repeatedly:

}
```
Leads as output at serial using ATmega2560 Meaga to
1.04
but on chipKIT MAX32 to

1.40000

i tried to work arround useing the dtostrf function, but unfortunately the same occurance.

please try to simulate and may confirm or to reject

JacobChrist:
Nice find. Do you know if correct expected output is documented anywhere on the Arduino.cc? I have not attempted to confirm this yet.

majenkotech:
`dtostrf` is an avr-libc specific (non-standard) function. We emulate it (or try to), but it uses the same code as is used by `String`. (actually, `String` just calls `dtostrf`).

That simply uses `snprintf` to format the number into a string. 

`Print::print`, on the other hand uses its own internal `printFloat()` function. Maybe we should replace our `dtostrf` function with something based around that function for consistent results?

majenkotech:
Actually, looking at it, it's not quite that simple. I mis-read your values - 1.4000 != 1.04...!

So this looks like the formatting isn't working, and is dropping leading zeros from the fractional part (it splits it into integer and fractional components then prints "%d.%d" of them).  Maybe that second `%d` should be `%0d`...

pgunzelmann:
Yes this is exactly why i opened the issue, i written an interpollation function and had wondering why it is not correct. Do to String funktion also seem to use the dtostrf  function it should be improved.

@JacobChrist as majenkotech mentioned the return values are incorect, indepent from Arduino.cc

JacobChrist:
Ah, I missed that. Formatting is not the issue.

EmbeddedMan:
It would be great to have a fix for this. Anybody have time to submit a PR?

majenkotech:
I'd like to port dtostrf from avr-libc if possible.

EmbeddedMan:
Very good. Let me know if you have a PR that needs testing and I'll help
out. Thanks for working on this Matt.

majenkotech:
Unfortunately, `dtostrf` relies heavily on `__ftoa_engine`, which is written in AVR assembly. So that's a bit of a non-starter really. Shame. It would have been a good way to get 100% compatibility. 

[...]

PR #432, opened 2018-07-01
Rewrite dtostrf to use snprintf with float formatting

majenkotech:
See issue #429


majenkotech:
I have just found a useful file in the ESP8266 core:

* https://github.com/esp8266/Arduino/blob/master/cores/esp8266/core_esp8266_noniso.c

Has everything we could ever want in it...

majenkotech:
There. No dodgy snprintf version any more. We now have a proper implementation lifted from ESP8266 (it's GPL'd, so that should be fine). We'll keep the original copyright intact.

PR #434, opened 2018-09-04
Fixed "pasting "/*JK*/" and "SET" does not give a valid preprocessing token" error

xlfdll:
The error occurred during compiling, when SD library was being used. Other libraries that include this file might also have such issue.

### DIFF
--- a/pic32/variants/Cerebot_MX4cK/Board_Defs.h
+++ b/pic32/variants/Cerebot_MX4cK/Board_Defs.h
@@ -399,7 +399,7 @@ extern const uint8_t	analog_pin_to_channel_PGM[];
 #define SD_CS_PIN 64
 
 //Pin 65
-#define	prtSDO				IOPORT_B	//JK
+#define	prtSDO				IOPORT_B
 #define	bnSDO				BIT_11
 
 //Pin 66


xlfdll:
Just for reference, the reproduce steps:

1. Set up Arduino environment to target to chipKIT Pro MX4 (MX4cK) board
2. Import library SD (or add #include <SD.h> at the beginning of the code)
3. Add code to use SD library, like
`pinMode(8, OUTPUT);`
`SD.begin(8);`
4. Compile.

The following error will occur:

> C:\Users\Xlfdll\AppData\Local\Arduino15\packages\chipKIT\hardware\pic32\2.0.5\variants\Cerebot_MX4cK/Board_Defs.h:402:31: error: pasting "/*JK*/" and "SET" does not give a valid preprocessing token
> 
>  #define prtSDO    IOPORT_B    //JK
> 
>                                ^

I reproduced this in 2 different PCs, one HP EliteBook 820 G3, and one customized-built PC. Both are with latest Windows 10 Enterprise. Arduino is installed using Windows installer method, and chipKIT package is installed using board manager (https://chipkit.net/wiki/index.php?title=How-To_Install_chipKIT_Core).

EmbeddedMan:
ok to test

Issue #435, opened 2018-09-09
servo library has inaccurate times on non 80MHz CPUs


EmbeddedMan:
Or at least that's how it appears. The following sketch does not generate the correct times on a FB Mini:

```
#include <Servo.h>

Servo THRout;

void setup() {
  THRout.attach(0);
}

void loop() {
  delay(1000);
  THRout.write(1000);
  delay(1000);
  THRout.write(1500);
  delay(1000);
  THRout.write(2000);
}
```

It does work on boards running at 80MHz. My guess is that the servo library does not correctly take into account FOsc, although that's just a guess at this point. 

Digging into the library now.

majenkotech:
```
#define usToTicks(_us)  (((_us)*5)/4)
#define ticksToUs(_ticks) ((((unsigned)(_ticks))*4)/5)
```

Looks hardcoded to me...

EmbeddedMan:
Exactly.

PR #436, opened 2018-09-09
Fis for Issue #435 : making servo library CPU speed aware

EmbeddedMan:
Added per-CPU speed #defines to make servo library take into account the speed and prescaler values of the timers it uses.


EmbeddedMan:
When looking at the diff, just look at commit #0ad60d6 (The others are just formatting cleanup)

PR #440, opened 2018-11-08
Implement USB.resume() for remote wakeup of PC

majenkotech:
Remote wakeup of the PC can now be achieved with `USB.resume();`.  It triggers the RESUME=1 / RESUME=0 sequence in the correct control register via the selected driver.


lstandage:
My testing has it working with Wi-Fire, WF32, and Lenny. 

I think we can merge.

EmbeddedMan:
Awesome! Thanks for testing it out Larry.

PR #443, opened 2018-12-16
Added required storage component for long filenames to actually work

majenkotech:
Although we had LFN turned on, it wasn't actually doing anything since it didn't have anywhere to store the long filename.


EmbeddedMan:
Looks good.

PR #447, opened 2019-03-12
Faster cdcacm

chromhelm:
Hi
This change enables the CDCACM to send with around 1M.
With -Os the buffer send should be at lest around 32byte-
With -O3 already with 4 bytes 1M is reached.
It is only reached when there is only one USB device registers in the USBManager.
It depents on PR #446

### DIFF
--- a/pic32/cores/pic32/USB_CDCACM.cpp
+++ b/pic32/cores/pic32/USB_CDCACM.cpp
@@ -209,7 +209,7 @@ bool CDCACM::onInPacket(uint8_t ep, uint8_t target, uint8_t __attribute__((unuse
             avail = CDCACM_BULKEP_SIZE;
         }
         
-        if((_txTail + avail) > CDCACM_BUFFER_SIZE)
+        if((_txTail + avail) >= CDCACM_BUFFER_SIZE)
 		{
 			uint8_t tbuf[avail];
 			
@@ -290,7 +290,7 @@ size_t CDCACM::write(uint8_t b) {
             avail = CDCACM_BULKEP_SIZE;
         }
 		
-		if((_txTail + avail) > CDCACM_BUFFER_SIZE)
+		if((_txTail + avail) >= CDCACM_BUFFER_SIZE)
 		{
 			uint8_t tbuf[avail];
 			
@@ -313,7 +313,7 @@ size_t CDCACM::write(uint8_t b) {
     return 1;
 }
 
-size_t CDCACM::write(uint8_t b, bool enqueuePacket) {
+inline size_t CDCACM::write(uint8_t b, bool enqueuePacket) {
 
     //if (_lineState == 0) return 0;
 
@@ -336,7 +336,7 @@ size_t CDCACM::write(uint8_t b, bool enqueuePacket) {
             avail = CDCACM_BULKEP_SIZE;
         }
 		
-		if((_txTail + avail) > CDCACM_BUFFER_SIZE)
+		if((_txTail + avail) >= CDCACM_BUFFER_SIZE)
 		{
 			uint8_t tbuf[avail];
 			
@@ -366,7 +366,7 @@ size_t CDCACM::write(const uint8_t *b, size_t len) {
     for (uint32_t i = 0; i < len; i++) {
 		bool equ = (i == len-1) || 										// enqueue if last byte
 					(i != 0 && i % CDCACM_BULKEP_SIZE == 0) ||				// enqueue on every full bulk buffer size
-					(((CDCACM_BUFFER_SIZE + _txHead - _txTail) % CDCACM_BUFFER_SIZE) < CDCACM_BUFFER_HIGH); // enqueue when the cdc x buffer is high
+					(((CDCACM_BUFFER_SIZE + _txHead - _txTail) % CDCACM_BUFFER_SIZE) < CDCACM_BUFFER_HIGH); // enqueue when the cdc tx buffer is high
 		write(b[i], equ);
     }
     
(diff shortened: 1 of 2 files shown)


pic32/cores/pic32/USB_CDCACM.cpp
```diff
@@ -263,8 +271,8 @@ size_t CDCACM::write(uint8_t b) {
 
     if (_lineState == 0) return 0;
 
-    volatile uint32_t h = _txHead;
-    volatile uint32_t newhead = (h + 1) % CDCACM_BUFFER_SIZE;
```
chromhelm: Is this really needed?

EmbeddedMan:
ok to test

majenkotech:
This looks good to me. About 850k/s with two CDCACMs loaded going to Minicom.

Vote to merge.

majenkotech:
Hold up - on the Lenny I'm seeing some data loss.  Using Paul S's USB test code I get about 550k/s on a WiFire. But on the Lenny I see this:

```
USB Fast Serial Transmit Bandwidth Test, capture this text.
USB Fast Serial Transmit Bandwidth Test, capture this text.
USB Fast Serial Transmit Bandwidth Test, capture this text.
USB Fas Serial Transmit Bandwidth Test, capture this text.
USB Fast Serial Transmit Bandwidth Test, capture this text.
USB Fast Serial Transmit Bandwidth Test, capture this text.
USB Fast Serial Transmit Bandwidth Test, capture this text.
USB Fast Serial Transmit Bandwidth Test, capture this text.
USB Fast Serial Transmit Bandwidth Test, capture this text.
```

Every so often there's a letter missing from the data.

**Hold on merge for now**

EmbeddedMan:
OK, good to know. We'll hold until the data loss issue is solved.

majenkotech:
The test code used:

```
// USB Serial Transmit Bandwidth Test
// Written by Paul Stoffregen, paul@pjrc.com
// This benchmark code is in the public domain.
//
// Within 5 seconds of opening the port, this program
// will send a message as rapidly as possible, for 10 seconds.
//
// To run this benchmark test, use serial_read.exe (Windows) or
// serial_listen (Mac, Linux) program can read the data efficiently
// without saving it.
// http://www.pjrc.com/teensy/serial_listen.c
// http://www.pjrc.com/teensy/serial_read.c
// http://www.pjrc.com/teensy/serial_read.exe
//
// You can also run a terminal emulator and select the option
// to capture all text to a file.  However, some terminal emulators
// may limit the speed, depending upon how they update the screen
// and how efficiently their code processes the imcoming data.  The
// Arduino Serial Monitor is particularly slow.  Only use it to
// verify this sketch works.  For actual benchmarks, use the
// efficient receive tests above.
//
// Full disclosure: Paul is the author of Teensyduino. 
//
// Results can vary depending on the number of other USB devices
// connected.  For fastest results, disconnect all others.


//#define USBSERIAL Serial       // for Leonardo, Teensy, Fubarino
#define USBSERIAL SerialUSB  // for Due, Maple

USBFS usbDevice;
USBManager USB(usbDevice, 0xf055, 0x1234);
CDCACM SerialUSB;
void setup()
{
  USB.addDevice(SerialUSB);
  USB.begin();
  USBSERIAL.begin(115200);
}

void loop()
{
  // wait for serial port to be opened
  while (!USBSERIAL) ;

  // give the user 5 seconds to enable text capture in their
  // terminal emulator, or do whatever to get ready
  for (int n=5; n; n--) {
    USBSERIAL.print("10 second speed test begins in ");
    USBSERIAL.print(n);
    USBSERIAL.println(" seconds.");
    if (!USBSERIAL) break;
    delay(1000);
  }

  // send a string as fast as possible, for 10 seconds
  unsigned long beginMillis = millis();
  do {
    USBSERIAL.print("USB Fast Serial Transmit Bandwidth Test, capture this text.\r\n");
  } while (millis() - beginMillis < 10000);
  USBSERIAL.println("done!");

  // after the test, wait forever doing nothing,
  // well, at least until the terminal emulator quits
  while (USBSERIAL) ;
}
```

majenkotech:
The problem possibly stems from #446 as I am getting random "1" inserted with this test code on that PR.

pic32/cores/pic32/USB_CDCACM.cpp
```diff
@@ -281,14 +289,69 @@ size_t CDCACM::write(uint8_t b) {
...
-            tbuf[i] = _txBuffer[_txTail];
+    if (enqueuePacket && _manager->canEnqueuePacket(_epBulk)) {
+        uint32_t s = disableInterrupts();
+        uint32_t avail = (CDCACM_BUFFER_SIZE + _txHead - _txTail) % CDCACM_BUFFER_SIZE;
+        if (avail > CDCACM_BULKEP_SIZE) {
+            avail = CDCACM_BULKEP_SIZE;
         }
-        _manager->enqueuePacket(_epBulk, tbuf, avail);
+		
+		if((_txTail + avail) > CDCACM_BUFFER_SIZE)
```
majenkotech: Shouldn't this be `_txTail + avail + 1`?

pic32/cores/pic32/USB_CDCACM.cpp
```diff
@@ -281,14 +289,69 @@ size_t CDCACM::write(uint8_t b) {
         if (avail > CDCACM_BULKEP_SIZE) {
             avail = CDCACM_BULKEP_SIZE;
         }
+		
+		if((_txTail + avail) > CDCACM_BUFFER_SIZE)
```
majenkotech: Same here?

chromhelm: I will test that in the afternoon. 
It can be implemented as `_txTail + avail >= CDCACM_BUFFER_SIZE`

majenkotech: Or `> CDCACM_BUFFER_SIZE - 1`... many ways to skin that cat :)

majenkotech:
Yep, that's fixed it.  I think we're good to go with this PR now.

PR #450, opened 2019-03-14
Added SDVOL pullup control and definition

majenkotech:
Also fixed the chip select pin number for the WiFire.

The theory here is quite simple:

* SDVOL was changed some time back to take a chip select pin instead of a pullup pin (for obvious reasons)
* No facility for controlling the pullup pin was retained
* This puts that facility back in using a simple #define in the board definition


majenkotech:
Also fixed the WF32 definitions as well.

PR #453, opened 2019-05-24
Fix D8 and D9 for Olimex Pinguino32 OTG board, Rev. D

CyberCastle:
And add other fixes from here: https://github.com/OLIMEX/Arduino_configurations/tree/master/PIC


EmbeddedMan:
Ok to test

EmbeddedMan:
I don't have a good way to test this out, but you only made changes in the Pinguino board files, so I'll go ahead and do the merge. Many many thanks for contributing your changes back to chipKIT!

PR #454, opened 2019-08-12
Fix spelling of __PIC32_HAS_L1CACHE

majenkotech:
The macro `__PIC32_HAS_L1_CACHE` in cpp-startup.S should actually read `__PIC32_HAS_L1CACHE`.

With the mis-spelling of the macro the core never turned on the L1 cache, instead relying on the bootloader to do it for it. Unless the chipKIT STK500 bootloader was being used (less and less the case with MZ chips now) the cache was never being enabled.

### DIFF
--- a/pic32/cores/pic32/cpp-startup.S
+++ b/pic32/cores/pic32/cpp-startup.S
@@ -230,7 +230,7 @@ _no_nmi:
         nop
 #endif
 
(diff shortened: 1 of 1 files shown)


EmbeddedMan:
This is such a fantastic find Matt. Nice job. Thank you so much for spending your time digging into this and getting to the bottom of it.

Issue #240, opened 2016-07-29
Wire.requestFrom(address, quantity, stop) missing in chipKIT-core 1.2.1


JacobChrist:
As of Arduino 1.0.1, requestFrom() accepts a boolean argument changing its behavior for compatibility with certain I2C devices.

If true, requestFrom() sends a stop message after the request, releasing the I2C bus.

If false, requestFrom() sends a restart message after the request. The bus will not be released, which prevents another master device from requesting between messages. This allows one master device to send multiple requests while in control.

The default value is true.

see: https://www.arduino.cc/en/Reference/WireRequestFrom

bjimmy77:
Trying to use some libraries that reference Wire.requestFrom. It is expecting to be able to use 3 argumnets, but the chipKit Wire.h only supports 2. It seems the chipKit Wire library is out of date.

JacobChrist:
> Trying to use some libraries that reference Wire.requestFrom. It is expecting to be able to use 3 argumnets, but the chipKit Wire.h only supports 2. It seems the chipKit Wire library is out of date.

Do you have time to update it, that would be great?

majenkotech:
My initial thoughts are that it would be as simple as making [this line](https://github.com/chipKIT32/chipKIT-core/blob/master/pic32/libraries/Wire/Wire.cpp#L204) optional. DTWI is a bit of a black box to me though.

PR #459, opened 2020-05-13
#240 Made Wire receiveFrom stopMaster() optional

majenkotech:
This adds an optional boolean parameter to `Wire.requestFrom()` which then is used to wrap the `di2c.stopMaster()` call to make executing it optional.

### DIFF
--- a/pic32/libraries/Wire/Wire.cpp
+++ b/pic32/libraries/Wire/Wire.cpp
@@ -182,7 +182,7 @@ void TwoWire::begin(int address)
     begin((uint8_t)address);
 }
 
-uint8_t TwoWire::requestFrom(uint8_t address, uint8_t quantity)
+uint8_t TwoWire::requestFrom(uint8_t address, uint8_t quantity, bool sendstop)
 {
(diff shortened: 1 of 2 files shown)


majenkotech:
# BEFORE MERGING THIS NEEDS TESTING BY SOMEONE WITH A SUITABLE SETUP

I don't have a setup that needs this functionality at the moment and can't test the knock-on effect of it. If someone that has a setup that would use this parameter could test it before merging that would be a huge benefit.

majenkotech:
Potential fix for issue #240

EmbeddedMan:
What is required to test that this change works? Generating a request to an I2C slave in multiple parts without a stop in between each requestFrom() except the last?

majenkotech:
Yeah, with some device that expects to work that way.

EmbeddedMan:
If such a device can be identified, I may be able to test.

majenkotech:
I have a feeling things like EEPROMS may work this way. First set an address, then read from that address without restarting.

PR #462, opened 2020-09-10
WIP: Add GitHub actions

adamwolf:
I set up Github Actions.

The file is in `.github/workflows/ant.ci`.  I am not an expert at GitHub Actions, but I think this replicates what we were doing with the core before on the Wayne and Layne Jenkins box.

GitHub expires the artifacts after a bit, so we'll still need to use Releases for long term archiving, I think.

(There are fun things we can do with this in the future too, like https://github.com/actions/create-release.)

### DIFF
--- a/.github/workflows/ant.yml
+++ b/.github/workflows/ant.yml
@@ -1,0 +1,65 @@
+name: Ant CI
+
+on: [push]
+
+jobs:
+  build:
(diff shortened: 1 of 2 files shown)


adamwolf:
You can see the Github Actions page for this commit on my fork: https://github.com/adamwolf/chipKIT-core/actions/runs/248331778

PR #483, opened 2022-08-27
Update Wire.cpp

ElectronicsSam:
Bugfix, when slave NACK received, next transmission would freeze system due to transmit buffer not empty from previous transmission.

This can happen when a slave disappears from the bus. In my example the SCL would stay low and my clicker 2 board would freeze completely due to while loop in beginTransmission that blocks until transmit buffer is empty.
Wire freeze when slave disappears
Wire one packet before


JacobChrist:
I think I have personally experienced this.  What device are you taking to and what board are you using?  Fubarino Mini 2.0 and PONTECH Quick 240's are easy boards for me to test with.

Without digging deeper the change look innocuous none the less.

